PR #1, opened 2020-05-22
Fix UDMP detection

bdraco:
The X-CSRF-Token header would be uppercase so
a get is needed in order for it to be found

Remove references to urllib3 since aiohttp
is used now instead.


briis:
Thanks @bdraco Appreciate your help. This has now been merged.

PR #2, opened 2020-06-13
Only update cameras detail every 60s (Keep events at every 2s)

bdraco:
My protect server saw a cpu spike and was a few
degrees warmer after installing the unifiprotect
integration due to the frequent polling.  When there
are >10 cameras this problem is much worse.

The bootstrap api is an expensive api to run
every 2s, which is the current update interval.
The calls to bootstrap are enough to have
a measurable impact on the event loop

The events api is rather cheap so its not a concern.

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -8,6 +8,18 @@ import jwt
 import aiohttp
 from aiohttp import client_exceptions
 
+CAMERA_UPDATE_INTERVAL_SECONDS = 60
+
+EMPTY_EVENT = {
+    "event_start": None,
+    "event_score": 0,
+    "event_thumbnail": None,
+    "event_on": False,
+    "event_ring_on": False,
+    "event_type": None,
+    "event_length": 0,
+}
+
 
 class Invalid(Exception):
     """Invalid return from Authorization Request."""
@@ -53,6 +65,7 @@ class UpvServer:
         self.is_unifi_os = None
         self.api_path = "api"
         self._is_authenticated = False
+        self._last_camera_update_time = 0
         self.access_key = None
         self.device_data = {}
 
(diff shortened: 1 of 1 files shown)


bdraco:
I wasn't 100% sure if the `last_ring` and `last_motion` should have been the start or end time.

bdraco:
Before (all data coordinator updates):
pyunifiprotect is most of my instances entire data update coordinator overhead.
Screen Shot 2020-06-13 at 10 25 32 AM

After (all data coordinator updates):
pyunifiprotect doesn't show up in the data update coordinator overhead.
Screen Shot 2020-06-13 at 10 25 59 AM

briis:
Hi @bdraco 
Thanks for the PR, I "only" have 4 Cameras so I did not notice that. But I do agree that there is information we don't need every 2 seconds. I will upload this as version 0.16 to PyPi, but I need the only guy I know with a Doorbell to test it before I change the manifest on the Unifi Protect Integration.

briis:
Another thought: Isn't there a way in Python to create an indefinite loop, that monitors the event stream in the background, and then pushes the information back to Home Assistant? Then we could just set the Pull frequency to 60 seconds for the Camera data and listen for the Events?

I am not a skilled Python programmer, everything I can do, I learned from guys like you, and watching what others have done, so I would have no idea where to start, but I would guess it would be some kind of Callback?

bdraco:
> Another thought: Isn't there a way in Python to create an indefinite loop, that monitors the event stream in the background, and then pushes the information back to Home Assistant? Then we could just set the Pull frequency to 60 seconds for the Camera data and listen for the Events?
> 
> I am not a skilled Python programmer, everything I can do, I learned from guys like you, and watching what others have done, so I would have no idea where to start, but I would guess it would be some kind of Callback?

If the protect server can deliver events over a websocket, a callback can be used to get them i near real-time.  I know the unifi controller has some form of websocket support but I'm not that familiar with the protect api.

briis:
Thanks - I don't know either, but I could poke around to see. Do you know of a good example where this is implemented? Then I could play with it, if I find that Protect is delivering data.

bdraco:
> Thanks - I don't know either, but I could poke around to see. Do you know of a good example where this is implemented? Then I could play with it, if I find that Protect is delivering data.
> 

Here is an example of websocket use with the ambient weather integration.
https://github.com/bachya/aioambient/blob/dev/aioambient/websocket.py

> Btw: If it was not clear, the V0.16, with your changes, is uploaded to PyPi, so you can change your manifest already now to use that version.

Thanks.  I had manually patched it :)

briis:
> Here is an example of websocket use with the ambient weather integration.
https://github.com/bachya/aioambient/blob/dev/aioambient/websocket.py

Thanks. I will see if I can figure something out in the next weeks.

PR #3, opened 2020-08-08
added websocket connection for motion events

adrum:
Initial Support for a WebSocket connection. It looks like only motion events are available for this interface.

It looks like an additional WebSocket connection is used by the Unifi Protect web interface that may expose doorbell rings, but the messages are binary, so this will have to suffice for now.

Things I'm not sure about:
- WebSocket connection lifecycle: I've not tested unexpected disconnects, timeouts, etc., and recovering from those events.
- Are changes required in https://github.com/briis/unifiprotect to take advantage of the new real-time updates for motion events?

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -98,7 +98,7 @@ class UpvServer:
         self._reset_camera_events()
         await self._get_events(10)
 
-        if self.ws is None:
+        if self.ws is None and self.is_unifi_os == True:
             if self.ws_task is not None:
                 try:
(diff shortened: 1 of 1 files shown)


briis:
Thanks for this Austin - I am still away from home on vacation, so cannot really test it the next few days, but looks great.
And yes, for the Unifi Protect Integration to benefit from this, we need to create some kind of listener in the integration, I have seen other do this with a callback, but I am not completely sure how to set this up. If you know more about this, maybe you can give it a shot?

briis:
If you want to give it a try, you can reference pyunifiprotect==0.19.0 that will contain your WS addon

PR #5, opened 2020-10-23
use nvr host for rtsp stream

adrum:
I'm not sure what causes this, but one of my cameras was using my public IP instead of the NVR hostname/ip for RTSP streams. This ensures the proper ip/hostname is set when streaming via RTSP by setting it to the NVR.

I don't believe there to be any negative side effects of this change, as I believe all RTSP streams are routed through the NVR and not direct to the camera.

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -309,7 +309,7 @@ class UpvServer:
                         channels = camera["channels"]
                         for channel in channels:
                             if channel["isRtspEnabled"]:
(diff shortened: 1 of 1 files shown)


briis:
Agree. Users have reported a few instances where this happens, so this is a good solution.

adrum:
Excellent! Glad to hear I wasn't the only one. Any chance you can add the label `hacktoberfest-accepted`?

briis:
Label is added. I forgot to give you credit for this, in the release notes. That will be added also.

PR #10, opened 2020-11-02
Ensure doorbell events are not missed when using websockets

bdraco:
If a motion event happened between when the doorbell was
rung and the events were fetched, the doorbell would
not be triggered


briis:
Wonderfull, thanks.

PR #11, opened 2020-11-16
Construct events from websocket updates

bdraco:
In the original version, websocket events caused a poll to the
events api.  The events are now fully constructed from the websocket
add/update messages.

Fixes https://github.com/briis/unifiprotect/issues/139

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,113 @@
+[tool.black]
+target-version = ["py37", "py38"]
+exclude = 'generated'
+
+[tool.isort]
+# https://github.com/PyCQA/isort/wiki/isort-Settings
+profile = "black"
+# will group `import x` and `from x import` of the same module.
+force_sort_within_sections = true
+known_first_party = [
+    "pyunifiprotect",
+]
+combine_as_imports = true
+
+[tool.pylint.MASTER]
+ignore = [
+    "tests",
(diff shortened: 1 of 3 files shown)


briis:
Thanks for starting the work on this. I pushed 0.24.0 to Pypi. A quick test shows that it detects motion when Recording mode is not never, but not when recording mode is set to never. Need to test person detection, which I will do later.

briis:
Sometimes after motion has been detected, the binary sensor is not resetting, so it stays in the on position. One scenario that triggers this behaviour is when motion recording is enabled, when the Integration Loads. Then triggering motion on the camera, will trigger the motion and then it resets correctly. 

Person detection seems to work perfectly.

bdraco:
> Sometimes after motion has been detected, the binary sensor is not resetting, so it stays in the on position. One scenario that triggers this behaviour is when motion recording is enabled, when the Integration Loads. Then triggering motion on the camera, will trigger the motion and then it resets correctly.
> 
> Person detection seems to work perfectly.

We are probably getting the initial event before the websocket connects and then it doesn't turn off because we get a partial event update that turns it off and we don't have the original camera id to know which camera it was for.

Probably can fix this by connecting the websocket before the event get and then skipping the event initial fetch when the websocket connect is in progress.

PR #12, opened 2020-11-16
Skip event get if websocket is reconnecting

bdraco:
I couldn't replicate the issue in https://github.com/briis/pyunifiprotect/pull/11#issuecomment-727761135
However if the problem is what I think it is, this should fix the event not being reset on the first event after initial load.

For testing:

I left the events viewer up at in the protect web ui.  I verified the event cleared
when the event finished in the events view in the protect web ui.

Verified this works when restarting while an event is happening.

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -103,15 +103,6 @@ class UpvServer:  # pylint: disable=too-many-public-methods, too-many-instance-a
         else:
             _LOGGER.debug("Skipping camera update")
 
-        # If the websocket is connected
-        # we do not need to get events
-        if self.ws_connection:
-            _LOGGER.debug("Skipping update since websocket is active")
-            return self.devices if camera_update else {}
-
-        self._reset_camera_events()
-        updates = await self._get_events(lookback=10)
-
         if (
             self.is_unifi_os
             and (current_time - WEBSOCKET_CHECK_INTERVAL_SECONDS)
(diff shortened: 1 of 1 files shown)


bdraco:
I've also verified the duration / event lengths match up and as soon as the protect UI show the event is no longer in process the binary sensor clears.

I've been doing this by walking around with the laptop with the screen open like this:

Screen_Shot_2020-11-15_at_8_50_48_PM

briis:
I will give it a test run right away.

briis:
This removed the motion state being on, after restart. And it seems to trigger motion as expected, including person detection.
However, it was a bit unstable, but might just be my Mac not having sufficient WiFi coverage. (All my G4 Cameras are outside)
I have a few Team Meetings the next 3 hours, but will test more once that is over.

One more question: Now that you did this, would it be possible to trigger motion even if Recording Mode is Off?

bdraco:
> One more question: Now that you did this, would it be possible to trigger motion even if Recording Mode is Off?

I did some testing and it appears that events do not happen when when recording mode is off.  If they don't appear in events in the protect UI, I don't think it would be possible.

I only tested with the doorbell though as its too dark to test with any of the cameras.

I'll check back in tomorrow late afternoon after I sleep+meetings in the morning.  I'm on HST right now.

briis:
I try and go through the Homebridge code, as he has managed to get this to work. I'll see if I can find what he is looking at to get this to work.
Sleep well. (My day just started)

bdraco:
> I try and go through the Homebridge code, as he has managed to get this to work. I'll see if I can find what he is looking at to get this to work.
> Sleep well. (My day just started)

That probably means that the `camera` model key still registers motion events when the  recording is off even if they don't show up in the events ui.

We could use those when we see recording is turned off.  We probably need to implement the `camera` updates via websocket first so we know when to switch over since `camera` updates are still polling otherwise there would be a short period where we are listening for both `events` and `camera` updates which would likely have undefined behavior.

g'night

bdraco:
Screen Shot 2020-11-16 at 1 53 17 PM

I've found one that got stuck in the ON state.  I have the logs saved to `camera_stuck_on.log` so I'll hopefully have enough to track down why.

bdraco:
I got the motion detection working without recording. 


Right now the code is a bit messy.  I'm hoping to have some time later this week to clean it up.   I've got to implement the camera model updates first and then know to switch to the camera model when recording is turned off.

bdraco:
Also as suspected, the score will always be 0 if its not recording because the UDMP/CloudKey is apparently what processes and scores the motion.

briis:
Sounds great. Let me know when there is some stuff to test.

PR #13, opened 2020-11-18
Update cameras model live from the websocket

bdraco:
Work in progress

First step to support updating the camera data live so we can know when to switch from using events to the camera model when recording is turned off to see motion.

- [x] To test: verify all the switches still work

### DIFF
--- a/pyunifiprotect/unifi_data.py
+++ b/pyunifiprotect/unifi_data.py
@@ -30,6 +30,25 @@ MAX_EVENT_HISTORY_IN_STATE_MACHINE = MAX_SUPPORTED_CAMERAS * 2
 
 LIVE_RING_FROM_WEBSOCKET = -1
 
+CAMERA_KEYS = {
+    "state",
+    "recordingSettings",
+    "ispSettings",
+    "ledSettings",
+    "upSince",
+    "firmwareVersion",
+    "featureFlags",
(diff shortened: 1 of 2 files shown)


bdraco:
Tested and everything is working as expected except for highFps switches.

However it seems to be an existing issue as the the api call to switch it isn't actually working.

briis:
One question first: Whats your first name?  Mine is Bjarne.
The highFps worked a while ago, but they might have changed the API.

Is this ready to merge for me testing it?

bdraco:
> One question first: Whats your first name? Mine is Bjarne.

My first name is actually `John`, but I use my middle name, `Nick` since so many people in my family are named `John` and it gets a bit confusing.  My github shows as `J. Nick Koston`

> The highFps worked a while ago, but they might have changed the API.
> 
> Is this ready to merge for me testing it?

Yes!

briis:
I'll call you Nick then. 👍 

I'll merge and start testing.

briis:
Everything seems to work very well. It is night here, so I will test a bit more in daylight. (All my G4's are outside)
With regards to highFps. I can get it to work partially: I can turn it off from HA but I cannot turn it back on. Switch state is also updated correctly, when I turn it on or off from the Protect App. So it just seems that the ON command is wrong.

PR #15, opened 2020-11-28
Always turn off events on end

bdraco:
Fixes https://github.com/briis/unifiprotect/issues/146


briis:
Thanks Nick. Merged, and will be v0.24.5

PR #16, opened 2020-12-02
Explictly check for ring events

bdraco:


bdraco:
Tested with homekit as well now.

Should be good to go.

briis:
Thanks Nick.

PR #17, opened 2020-12-18
Ignore cameras that are not adopted to the controller

bdraco:
The api in 1.17+ now shows non-adopted cameras / cameras on from other controller.

Note: I had to delete the extra "None" entities manually.

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -315,6 +315,12 @@ class UpvServer:  # pylint: disable=too-many-public-methods, too-many-instance-a
         if not self.ws_connection and "lastUpdateId" in json_response:
             self.last_update_id = json_response["lastUpdateId"]
         for camera in json_response["cameras"]:
+
+            # Ignore cameras adopted by another controller on the same network
+            # since they appear in the api on 1.17+
+            if "isAdopted" in camera and not camera["isAdopted"]:
+                continue
+
             camera_id = camera["id"]
 
             first_update = camera_id not in self.device_data
@@ -533,8 +539,8 @@ class UpvServer:  # pylint: disable=too-many-public-methods, too-many-instance-a
         """Sets the camera recoding mode to what is supplied with 'mode'.
         Valid inputs for mode: never, motion, always, smartDetect
         """
-        if 'smart' in mode:
-            mode = 'smartDetect'
+        if "smart" in mode:
+            mode = "smartDetect"
 
         await self.ensure_authenticated()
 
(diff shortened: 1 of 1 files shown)


briis:
This fixed the issue with the camera.none, but I get a new error now:

```python
  File "/usr/lib/python3.8/site-packages/pyunifiprotect/unifi_protect_server.py", line 871, in _setup_websocket
    self._process_ws_message(msg)
  File "/usr/lib/python3.8/site-packages/pyunifiprotect/unifi_protect_server.py", line 931, in _process_ws_message
    self._process_camera_ws_message(action_json, data_json)
  File "/usr/lib/python3.8/site-packages/pyunifiprotect/unifi_protect_server.py", line 938, in _process_camera_ws_message
    camera_id, processed_camera = camera_update_from_ws_frames(
  File "/usr/lib/python3.8/site-packages/pyunifiprotect/unifi_data.py", line 262, in camera_update_from_ws_frames
    processed_camera = process_camera(None, host, camera, True)
  File "/usr/lib/python3.8/site-packages/pyunifiprotect/unifi_data.py", line 100, in process_camera
    online = camera["state"] == "CONNECTED"
KeyError: 'state'
```
I have tried to turn off my CloudKey, just to test if that had anything to do with it, but that still introduces this error.

briis:
Hi Nick,
I believe I fixed the new error I got above.

I have added the following to unifi_protect_server.py in line 942:
```python
        if camera_id is None or processed_camera is None:
```
instead of 
```python
        if camera_id is None
```

and then I have added the below block to unfi_data.py in the *process_camera* function:
```python
    # If Non Adopted Camera, abort
    if "state" not in camera:
        return
```

It does work, but is this the right way to do it?

bdraco:
> Hi Nick,
> 
> I believe I fixed the new error I got above.
> 
> 
> 
> I have added the following to unifi_protect_server.py in line 942:
> 
> ```python
> 
>         if camera_id is None or processed_camera is None:
> 
>            return
> 
> ```
> 
> instead of 
> 
> ```python
> 
>         if camera_id is None
> 
>            return
> 
> ```
> 
> 
> 
> and then I have added the below block to unfi_data.py in the *process_camera* function:
> 
> ```python
> 
>     # If Non Adopted Camera, abort
> 
>     if "state" not in camera:
> 
>         return
> 
> ```
> 
> 
> 
> It does work, but is this the right way to do it?
> 
> 


Nice.  Looks like we need to filter the extras cameras from getting web socket updates.

Looks good to me as long as "state" will always be provided for adopted cameras.  I assume it will be.

It might be a bit cleaner to abort if the camera id isn't in the dict of known cameras.

briis:
Please feel free, to make the RIGHT change - I enjoy having you onboard here as I learn a lot. So I prefer the clean way, that just something that works.

PR #18, opened 2020-12-19
Skip cameras that are not present in the state machine

bdraco:
I added `has_camera` to the state machine so we can skip cameras that aren't present.

### DIFF
--- a/pyunifiprotect/unifi_data.py
+++ b/pyunifiprotect/unifi_data.py
@@ -20,8 +20,8 @@ EVENT_LENGTH_PRECISION = 3
 
 TYPE_RECORD_NEVER = "never"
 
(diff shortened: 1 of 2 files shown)


bdraco:
Running this for a bit to make sure that it still addresses the error before I mark it ready for review.

I didn't wait long enough on the previous PR. The non-adopted camera updates are very infrequent for my test setup so I want to make sure this runs for at least an hour without error before I mark it ready

briis:
Yes, I saw the same when I tried to fix it this afternoon - It took a little while before the errors started.
You just push it, when it's ready.

bdraco:
Looks good after running for an hour 👍

briis:
Perfect, thanks. 
I'll merge it. 

Have a good weekend.

PR #20, opened 2021-01-03
Moved to device vs. camera

briis:
Still missing the Services to change things on the Floodlight, but wanted to agree on structure first.

### DIFF
--- a/pyunifiprotect/unifi_data.py
+++ b/pyunifiprotect/unifi_data.py
@@ -61,9 +61,19 @@ CAMERA_KEYS = {
     "lastRing",
     "isMotionDetected",
     "zoomPosition",
-    "light_mode",
 }
 
+LIGHT_KEYS = {
+    "isConnected",
+    "name",
+    "type",
+    "light_mode",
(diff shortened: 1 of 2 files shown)


bdraco (approved):
LGTM

bdraco:
Is there an option to disable/enable motion detection for the light in the UI?  If not we can probably remove the code to generate events from the `light` model since we can likely assume there will always be events if there isn't an option.

briis:
No, motion detection by itself, cannot be disabled - it's like the cameras. You can however set the floodlight so that motion on the PIR turns on the light: always, when it is dark or disable it.

So your suggestion should work. I am not sure exactly what part of the code you would remove?

bdraco:
> No, motion detection by itself, cannot be disabled - it's like the cameras. You can however set the floodlight so that motion on the PIR turns on the light: always, when it is dark or disable it.
> 
> So your suggestion should work. I am not sure exactly what part of the code you would remove?

I implemented that in https://github.com/briis/pyunifiprotect/pull/21  I don't have a way of testing it though

PR #21, opened 2021-01-03
Simplify light processing

bdraco:
This assumes that events are always generated for light motion detection and we do not need to
create them from the light model like we do for cameras when motion detection is disabled.

### DIFF
--- a/pyunifiprotect/unifi_data.py
+++ b/pyunifiprotect/unifi_data.py
@@ -75,6 +75,7 @@ LIGHT_KEYS = {
     "lastMotion",
 }
 
+
 @enum.unique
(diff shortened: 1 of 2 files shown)


briis:
I rolled the changes back, as it turned out that when motion is set to manual, no motion events are generated. With this, it works even when manual is set.
This is the method I prefer as it is consistent with the way we do it for the Cameras.

bdraco:
> I rolled the changes back, as it turned out that when motion is set to manual, no motion events are generated. With this, it works even when manual is set.
> 
> This is the method I prefer as it is consistent with the way we do it for the Cameras.

Sounds good 👌

PR #22, opened 2021-02-01
Fix first update race condition

bdraco:
Should fix other cases of https://github.com/briis/unifiprotect/issues/181

Fix unique id not being unique

Run code though black and isort


briis:
Thanks.

PR #23, opened 2021-02-03
Setup a basic CI

bdraco:
https://github.com/bdraco/pyunifiprotect/actions/runs/534836772

We can give some basic feedback if there is a problem


briis:
Excuse my ignorance Nick, but what exactly does this do? I can see the linting and tests, so it is more the conceptual understanding I need. As you know I am self-taught, so I just try to understand what and why? So I can improve.

PR #24, opened 2021-02-03
Ensure websocket updates always contain the full processed data

bdraco:
When we called back the subscribers in fire_event we only
sent the changes. Sometimes this would be missing event_on
or event_ring_on if the websocket connected before the
integration setup the binary sensor. We now always return
the full processed data.

Should fix the last case of https://github.com/briis/unifiprotect/issues/181

Various cleanups (there were too many things called `data` to keep track of)

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -84,7 +84,7 @@ class UpvServer:  # pylint: disable=too-many-public-methods, too-many-instance-a
         self._last_device_update_time = 0
         self._last_websocket_check = 0
         self.access_key = None
-        self.device_data = {}
+        self._processed_data = {}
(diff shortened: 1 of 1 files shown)


bdraco:
Doing some double checking on my testing

bdraco:
Looks good

pyunifiprotect/unifi_protect_server.py
```diff
@@ -1137,8 +1141,8 @@ def fire_event(self, device_id, processed_event):
         self._update_device(device_id, processed_event)
 
         for subscriber in self._ws_subscriptions:
-            subscriber({device_id: processed_event})
+            subscriber({device_id: self._processed_data[device_id]})
```
bdraco, line 1144: This is the actual fix.

briis:
Looks great, thanks.

PR #26, opened 2021-03-21
feat(g4doorbell/chime_duration): make chime duration adjustable

mohlek:
proposal pr


briis:
Thanks Maurice,
I will upload it to pypi.org in a minute as 0.30.6

PR #27, opened 2021-05-01
 Only clear the event_object if smartDetectEvents is not set in the followup motion event

bdraco:
Maybe fixes https://github.com/briis/unifiprotect/issues/225


briis:
Thanks Nick. I am on a mini-vacation with the Wife, back in a day, so I will merge when I get back, so I can test it before releasing.

Again thanks a lot for your help.

PR #29, opened 2021-07-29
General project maintenance

AngellusMortis:
This PR is just some general cleanup/maintenance/recommendations for the overall structure of the project. **There should be no actual changes to the code base from this PR**.

I am completely open to removing/splitting discussing any of the changes. They are just a bunch of "guidelines" from personal/work projects I have worked on how I prefer to setup a python project.

* Adds [VS Code Devcontainer support](https://code.visualstudio.com/docs/remote/create-dev-container)

  For anyone that already has Docker/VS Code installed, this allows you to very rapidly get a fully configured development environment, including configured linters and integrate test running in VS Code with no extra setup. It requires Docker, VS Code and the [Remote Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension and then when you open the project folder in VS Code, you will get a prompt to reopen in a devcontainer.

  This would not replace users from using a virtualenv if they prefer or stop virtualenv dev setups from working.

* Removes duplicate README

* Updates to latest setuptools build backend (setup.py -> setup.cfg)

  If you follow the latest [setuptools packaging guide](https://setuptools.readthedocs.io/en/latest/setuptools.html), the `setup.py` has been replaced with just the `setup.cfg`. 1:1 migrate of `setup.py` to a `setup.cfg`

* Adds "dev" extra for all of the testing/linting dev requirements

  This allows you to do `pip install -e .[dev]` and you will get all of the expected testing/development requirements like tox, pylint, flake8, black, etc.

* Moved linting CLI parameters into config files so code editors pick them up

  There was some CLI parameters for flake8/pylint specified directly in the `tox.ini`. This moves them into their proper config files so code editors like VS Code will pick them up correctly. 

  Because flake8 does not yet support the project.toml file, I added the dev only requirement for [pyproject-flake8](https://pypi.org/project/pyproject-flake8/). It allows you to specific flake8 config options in the pyproject.toml file and you have to use the new cli entry point `pflake` instead of `flake8`.

* Fixes noqa in pyojectunifiproject/__init__.py

  Minor linting fix I could make in the __init__.py when I opened it. The `__all__` is how you specify to Python that these variables are "exportable". This also makes linters stop complaining about them not being used.

* Replaces requirements*.txt files with full pinned versions for repeatable builds

  A requirements*.txt file is suppose to be fully pinned and resolved. This so to ensure fully reproducible builds. If a new major version of tox is released between CI runs, it _could_ cause a repeat build to fail. So I replaced requirements_all.txt and requirements_test.txt with fully pinned/resolved versions of themselves generated via pip-tools.

  pip-tools, as opposed to pipenv or poetry, allows you easily maintain fully pinned/resolved requirements for CI/install while _not_ completely overhauling changing the Python workflow. It is my preferred solution to handling this problem as it does not completely change the normal setup process for Python developers. It is just a helper command (`pip-compile`) to resolve and create the pinned file.


AngellusMortis:
This is the first of at least 4 PRs I have planned to improve `pyunifiprotect` to make it more robust and allow it to pull more data from Unifi Protect so we can get more data into Home Assistant (media source, etc. whatever else).

If you would like to chat more about the project I have reddit, Signal, and Discord. I _should_ also probably really get around to making a Home Assistant Forums account as well.

The overall plan is:

1. Optional: Update project (this PR)
2. Improve CLI interface / add more tests -- this will let me ensure there is no backwards compatibility breaking changes going forward as well as allow us to quick verify a new version of Protect does not introduce breaking API changes.
3. Improve API Client / add typing + [Pydantic](https://pydantic-docs.helpmanual.io/) -- this is one of those that is "hard to explain unless you have seen the benefits", but I have that working with third APIs/code, it makes it a lot easier if you typing your data instead of using raw Javascript objects/JSON/Python dicts. It makes finding issues caused by unexpected data signfanctly easier to troubleshoot and diagnose. Pydantic is a really neat library because it essentially adds static typing to Python for "data models" so you can know _at runtime_ if there is bad data causing expected issues. I have done similar things in two other personal projects I have that interact with third party data: [Boundlexx](https://github.com/AngellusMortis/boundlexx) (no Pydantic, but it uses a database ORM and the database types/structures all of the data) and [sxm-player](https://github.com/AngellusMortis/sxm-player)
4. Build out data models / client methods for accessing more endpoints related to pulling media _out_ of Protect. Like /events/, /video/export, etc.

bdraco (changes requested):
This looks reasonable to me. 

Would you please make the change to `pyunifiprotect/__init__.py` in another PR?

bdraco:
> * Updates to latest setuptools build backend (setup.py -> setup.cfg)
>   If you follow the latest [setuptools packaging guide](https://setuptools.readthedocs.io/en/latest/setuptools.html), the `setup.py` has been replaced with just the `setup.cfg`. 1:1 migrate of `setup.py` to a `setup.cfg`


This might break the current publication workflow (even though its easy enough to use `python -m build` instead), briis will need to comment on this in its a problem.  For now it would be best to break that change into a new PR so he can have a look upon return from holiday.

AngellusMortis:
> Would you please make the change to pyunifiprotect/__init__.py in another PR?

I can if it is really necessary, but it is a noop change. All it does is remove the need for the linter override. `from pyunifiprotect import *` will still import the same items from the module as before since everything in the locals for that scope is put inside of the `__all__`.

briis:
Thank you for this. I am not completely sure about the implications of these changes, but I trust that you guys have a lot more experience than me. But to make sure I can continue to build things. This is how I publish a new release today:

```bash
% python3 setup.py sdist
% python3 -m twine upload dist/pyunifiprotect.VERSION.tar.gz
````

What command would I need to use after the changes?

**Comment by AngellusMortis** [comment]
Only the first command would change since the `setup.py` no longer exists. 

```
% python3 -m build
% python3 -m twine upload dist/pyunifiprotect.VERSION.tar.gz
```

I could also add this Github Workflow which will auto-publish to PyPi whenever you make a Github Release: https://github.com/actions/starter-workflows/blob/7135450ec14e7e47bf3e41dd65d399e5e315fa39/ci/python-publish.yml

bdraco:
> > Would you please make the change to pyunifiprotect/**init**.py in another PR?
> 
> I can if it is really necessary, but it is a noop change. All it does is remove the need for the linter override. `from pyunifiprotect import *` will still import the same items from the module as before since everything in the locals for that scope is put inside of the `__all__`.

The advantage to merging it now is it will save a bit of time as we will no longer have to approve and run github workflows since github will no longer treat you as a first time contributor as we work through this PR.

Its strait-forward change we can merge right away.

AngellusMortis:
https://github.com/briis/pyunifiprotect/pull/31

bdraco:
> Only the first command would change since the `setup.py` no longer exists.
> 
> ```
> % python3 -m build
> % python3 -m twine upload dist/pyunifiprotect.VERSION.tar.gz
> ```
> 
> I could also add this Github Workflow which will auto-publish to PyPi whenever you make a Github Release: [https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml](https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml?rgh-link-date=2021-08-01T12%3A17%3A04Z)


Thats sounds good to me. Personally I prefer `bump2version` which publishes on a new tag (https://github.com/bdraco/aiodiscover/blob/main/.github/workflows/build-main.yml#L58)

If you expect a large number of merges, setting up https://github.com/release-drafter/release-drafter might make sense as well to minimize the publication work on Bjarne's part.

bdraco:
I think this iteration is good to merge as-is once the publication documentation is added.  It doesn't need to be complex.  It should be the single source of truth on how to publish as things progress to make sure a release can always be done without having to figure out what has changed.

AngellusMortis:
I can remove the setup.py changes into another PR so we can have more discussion about releases processes and get more weight in from @briis on how he prefers it.

bdraco (approved):

PR #30, opened 2021-08-01
Allow requesting a custom snapshot width and height

bdraco:
Fixes https://github.com/briis/unifiprotect/issues/270


briis:
Thank you Nick. This one has been bugging me for a while, and then you even added the additional feature of specifying heigh and width.

bdraco:
Thanks.  I'm thinking ahead in the event we end up going this direction: https://github.com/home-assistant/architecture/discussions/595

PR #32, opened 2021-08-01
Update Package Publishing

AngellusMortis:
Split from previous PR: https://github.com/briis/pyunifiprotect/pull/29

Update to how package is created/published. @briis, just let me know which options you prefer and I can get the PR upload with docs for the correct one.

## Part 1 replacing `setup.py`

[PEP-517](https://www.python.org/dev/peps/pep-0517/) is finally standardizing how packages are build for Python. setuptools, the "tried and true" way of doing packages is recommending getting rid of the `setup.py` in favor of a `setup.cfg`. This is a great time to  look at alternatives as well.

### Option 1: Update to latest setuptools workflow (`setup.py` -> `setup.cfg`) -- current diff

This would change the `setup.py` to a `setup.cfg` and nothing else would really change.

To build publish package:

```bash
$ python3 -m build
$ python3 -m twine upload dist/pyunifiprotect.VERSION.tar.gz
````

To install in "development mode" it would still be the same

```bash
$ pip install -e ".[dev]"
```

### Option 2: Switch to using flit workflow (`setup.py` -> `pyproject.toml`)

If we want to clean up the extraneous files in project repo, we could switch [flit](https://flit.readthedocs.io/en/latest/), which would allow us to put the build metadata into the `pyproject.toml` file and all of the data about the project would exist in a single file.

This is what I am using in my projects because I really like the idea of only a single file. setuptools is the way most people are use to doing Python packaging so this is a bit different.

To build publish package:

```bash
$ flit build
$ flit upload
````

To install in "development mode":

```bash
$ pip install requirements_test.txt
$ flit install --symlink
```

## Part 2: Publishing to PyPi / creating releases

Each option is basically how much automation we want to have in the release process. Personally I would likely go with option 3 as I do think using bump2version has a lot of benefit when you already have Github Actions in place to help with automation. Thanks to @bdraco for showing me release-drafter, I think I need to add it to my other projects. I really like the idea of linking PyPi releases to Github releases. It makes everything nice and consistent. 

### Option 1: Manual -- current diff

We can keep the process 100% manual. 

Details: 

* Increase version numbers manually in code
* Manually run the build/upload commands above to create a release and publish it to PyPi

### Option 2: Automated Publish (`python-publish`)

We can use the [official Github Workflow template for publishing PyPi package](https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml).

Details:

* Increase version numbers manually in code
* Manually draft new release on Github
* Publish release on Github to automatically build/publish to PyPi

### Option 3: Release Drafter + Automated Publish

Same as 2 + adding in [release-drafer](https://github.com/release-drafter/release-drafter).

Details:

* Increase version numbers manually in code
* New release is drafted and updated automatically whenever a commit is added to `master`
* Publish release on Github to automatically build/publish to PyPi

### Option 4: bump2version + Release Drafter + Automated Publish

Same as 3 + adding in [bump2version](https://pypi.org/project/bump2version/) to automate creating tags and increases version numbers.

Details:

* Need to run `bump2version` command to increase version number (ex: `bump2version --current-version 0.5.1 minor src/VERSION`)
* New release is drafted and updated automatically whenever a commit is added to `master`
* Publish release on Github to automatically build/publish to PyPi


briis:
Hi both,
Thanks a lot for this, I like the idea of automatic publishing. I like both version 3 and 4, and I happy to follow your advice on what is best, as you are the experienced guys here.
I will merge this release now.

PR #34, opened 2021-08-21
Refactors CLI and how requests are made in `UpvServer`

AngellusMortis:
I planned to add more tests as part of this PR so we can more rapidly test for backwards compatibility issues / test new versions of Unifi Protect, but the refactor for how requests to Unifi Protect to make them easier to mock turned into a lot of changes and I did not want to make a bigger PR.

I can likely split this PR up into 3 smaller ones if requested. 

### General fixes

* Adds `url` to setuptools metedata
    * This will make the "Home Page" link appear in PyPi
* Fixes `black` formatting to respect 120 character limit (also had to re-run black on everything).
* Fixes `black` exclude to ignore standard python files
* Fixes `black` being missing from dev/test requirements
* Adds `.vscode/settings.json` to `.gitignore`. 
    * When using `devcontainers`, the `devcontainer.json` exists as the "base" settings and then the `settings.json` can be a user's own overrides. Sometimes VS Code may also create an empty file as well.

### CLI Refactor

* Adds `typer` and moves all loose `*.py` files into Typer subcommands
* Replaces `protectdata_function.py` with an interactive shell subcommand
    * Adds `ipython` and `termcolor` for this subcommand
    * Requirements are optional in a new extra: `shell` (this means HA will not pull them in as it does not need them)
* Adds `typer` entrypoint as to setuptools as `unifi-protect`
* Adds `python-dotenv` to load CLI args from env file
* Adds all of the commands as VS Code debug tasks so you can do interactive debugging in VS Code with them
* Updates `TESTDATA.md` to include new instructions using `unifi-protect` command

Screenshot of the new `shell` subcommand (with VS Code interactive debugging!):
image

### Request refactor

* Refactors how API requests are made in `UpvServer`
    * Adds `BaseApiClient` class and refactors/moves all request/auth related calls to it
    * Adds new `api_request` method that does many common things (error checking, decoding JSON)
    * Refactors all calls to Protect to use `api_request` to simplify methods and remove a lot of duplicated code.
* Fixes backwards compatibility breakage in https://github.com/briis/pyunifiprotect/pull/30. Even if an arg is optional, it needs a default.

### Notes for Testing

I successfully tested this with a Floodlight, a G4 Doorbell, a G4 Pro camera, and a G4 Bullet with the 0.8.9 version of `unifiprotect` HA Integration.

To test this, I went into my Home Assistant instance and edited `custom_components/unifiprotect/manifest.json` with following and then restarted Home Assistant. 

##### Old

```json
"requirements": [
    "pyunifiprotect==0.31.9"
],
```

##### New
```json
"requirements": [
    "git+https://github.com/AngellusMortis/pyunifiprotect.git@feature/cli-request-refactor#pyunifiprotect==0.31.9"
],
```


briis:
Thanks a lot Christopher. As there are so many changes  I have asked @bdraco also to look at this before I merge.

bdraco:
I'll take a look after I finish up everything for Home Assistant 2021.9 beta later this week

briis:
Thanks Nick.

AngellusMortis:
Yeah, the plan was to just change enough in `unifi_protect_server.py` so I could mock out all of the requests in a single place with a fixture, but that turned into a lot more work then I was expecting. 

With the as is in this PR, the plan is to add a new CLI to command to generate "sample data" for testing and then mock out `api_request` and probably `_setup_websocket`. Then I can write tests with 100% coverage for the rest of `UpvServer`.

pyunifiprotect/unifi_protect_server.py
```diff
@@ -65,131 +67,167 @@ class NvrError(Exception):
...
-    def devices(self):
-        """Returns a JSON formatted list of Devices."""
-        return self._processed_data
+    def is_unifi_os(self):
+        return bool(self._is_unifi_os)
 
-    async def update(self, force_camera_update=False) -> dict:
-        """Updates the status of devices."""
+    @is_unifi_os.setter
+    def is_unifi_os(self, value):
```
bdraco, line 114: Are there more changes coming here?

bdraco (commented):
Overall this looks good, but I only did a quick pass on my phone.  More feedback later this week.

pyunifiprotect/unifi_protect_server.py
AngellusMortis: Everything should be done done (for this PR). I have already tested in my HA instance and it works exactly the same, which is the goal. 

For this specific spot (`is_unifi_os`), this was just to get rid of the `Optional[Bool]` type on `is_unifi_os` to be honest. It should never actually be `None` since before it does anything else, `ensure_authenticated` would be called and `_is_unifi_os` would be set to a bool instead of `None`.

bdraco: Any reason to not keep it as a bare attribute and initialize it to False?

AngellusMortis: I thought about that, but then `check_unifi_os` would not be able to tell if it has been initialized. So if it was `False`, it would be forced to check if it is Unifi OS every time instead of just seeing it being `None` the first time and then checking once. 

I could refactor out the initialize authentication and check for Unifi OS into a `init` method, but it would require async since it makes HTTP requests so it cannot be in the class `__init__`.

bdraco: Ok to clean it up later 👍

bdraco:
Just tried this an got a trace
```
2021-08-22 14:44:57 ERROR (MainThread) [aiohttp.server] Error handling request
  File "/usr/local/lib/python3.8/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
    resp = await self._request_handler(request)
  File "/usr/local/lib/python3.8/site-packages/aiohttp/web_app.py", line 499, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.8/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 60, in security_filter_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/forwarded.py", line 89, in forwarded_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 24, in request_context_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 78, in ban_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 144, in auth_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/view.py", line 135, in handle
    result = await result
  File "/usr/src/homeassistant/homeassistant/components/camera/__init__.py", line 583, in get
    return await self.handle(request, camera)
  File "/usr/src/homeassistant/homeassistant/components/camera/__init__.py", line 601, in handle
    image = await _async_get_image(
  File "/usr/src/homeassistant/homeassistant/components/camera/__init__.py", line 164, in _async_get_image
    image_bytes = await camera.async_camera_image(
  File "/config/custom_components/unifiprotect/camera.py", line 333, in async_camera_image
    last_image = await self.upv_object.get_snapshot_image(
TypeError: get_snapshot_image() takes 2 positional arguments but 4 were given
```

bdraco:
Its likely this PR is missing the changes for 2021.9: <https://developers.home-assistant.io/docs/core/entity/camera/#camera-image>

https://github.com/briis/pyunifiprotect/commit/bf723be87f7759d7f06eb111950979eae06fab6c

AngellusMortis:
It definitely has the latest `master` for `pyunifiprotect` and it is definitely has the width/height args on `get_snapshot_image`:

https://github.com/briis/pyunifiprotect/pull/34/files#diff-deca4cd99e889cb04c9ad396127639506a3768d5ba5502975108ec493ec14266R614

bdraco:
It is there for sure. Probably Home Assistant downgraded the dep on me when I restarted.  Will give it another run

bdraco:
> It is there for sure. Probably Home Assistant downgraded the dep on me when I restarted. Will give it another run

Yup, that is what happened.  Sorry, bad testing on my part!

pyunifiprotect/unifi_protect_server.py
```diff
@@ -273,20 +419,8 @@ async def _get_unique_id(self) -> None:
     async def _get_server_info(self) -> None:
```
bdraco, line 419: Existing, but the return type needs to be fixed.

pyunifiprotect/unifi_protect_server.py
```diff
@@ -404,51 +522,34 @@ def _process_viewports_json(self, json_response, server_id, include_events):
...
 
     def _reset_device_events(self) -> None:
         """Reset device events between device updates."""
         for device_id in self._processed_data:
             self._update_device(device_id, PROCESSED_EVENT_EMPTY)
 
-    async def _get_events(
-        self, lookback: int = 86400, camera=None, start_time=None, end_time=None
-    ) -> None:
+    async def _get_events(self, lookback: int = 86400, camera=None, start_time=None, end_time=None) -> None:
```
bdraco, line 533: Return type needs to be fixed here.

pyunifiprotect/unifi_protect_server.py
```diff
@@ -464,135 +565,71 @@ async def _get_events(
...
     async def get_thumbnail(self, camera_id: str, width: int = 640) -> bytes:
         """Returns the last recorded Thumbnail, based on Camera ID."""
 
-        await self.ensure_authenticated()
         await self._get_events(camera=camera_id)
 
         thumbnail_id = self._processed_data[camera_id]["event_thumbnail"]
 
         if thumbnail_id is None:
             return None
```
bdraco, line 594: Return type is `bytes` but returning `None`

pyunifiprotect/unifi_protect_server.py
```diff
@@ -464,135 +565,71 @@ async def _get_events(
...
     async def get_heatmap(self, camera_id: str) -> bytes:
         """Returns the last recorded Heatmap, based on Camera ID."""
 
-        await self.ensure_authenticated()
         await self._get_events(camera=camera_id)
 
         heatmap_id = self._processed_data[camera_id]["event_heatmap"]
 
         if heatmap_id is None:
             return None
```
bdraco, line 610: Existing: Return type is `bytes` but returning `None`

bdraco:
Need to check that the websocket is still working as I'm seeing a few more requests to the event api then I expected

Screen Shot 2021-08-22 at 10 04 58 AM

AngellusMortis:
Nothing should have really changed with the Websocket. I changed how the URL was generated a bit to simplify it. The new `shell` command should make it super easy to debug though.

Also, as a side note, how do you generate that image?

bdraco:
> Also, as a side note, how do you generate that image?

https://www.home-assistant.io/integrations/profiler/#service-profilerstart

Call the `profiler.start` service, then download the callgrind.out.XXXX file and open it in qcachegrind

bdraco:
We should probably warn if we fallback to polling if unifi_os is True.

AngellusMortis:
It should be fixed up now. I made it so all of the logging is dumps to console for the `shell` command to make it easier to use for debugging. Also added the log statement if it falls back to polling.

bdraco:
Looks like just HomeKit snapshots now hitting the api so I assume the websocket is working now.  I'll do some more testing

Screen Shot 2021-08-22 at 10 35 56 AM

bdraco:
Definitely processing websocket frames

Screen Shot 2021-08-22 at 10 37 29 AM

bdraco:
Manual test of waving at a camera didn't trigger the motion sensor so something might still need to be adjusted

bdraco:
Tried another camera. Motion detection looks good.

AngellusMortis:
Re-installing the latest version now on my HA instance. To double check. I did noticed a couple of days ago that the motion was kind of iffy for just the whole "wave you hand around", but that is likely by design.

AngellusMortis:
Yep, motion looks good on my Bullet.

bdraco (approved):
Looks good.

Hopefully @briis can test as well before merging given the scope of the change since the test coverage on this project isn't going to tell us if there was something else that needs to be adjusted from the refactor.

briis:
Just about to have dinner, will test later this evening. Thank you both.

AngellusMortis:
I have most of the devices and tested with them. The only supported device I am missing is a Viewport since I have not gotten around to buying one yet. 

I did not know about the Websocket fall back though so I assumed it was all good when I could still do stuff.

briis:
I have been through testing, and found everything to be working, including the Viewport device, camera snapshot, motion detection etc. So I am happy to merge and and upload a new build to pypi.org. Any last changes?

AngellusMortis:
Nothing for me. I will make more PRs to prevent this one creeping any bigger.

briis:
Hi Guys,
There are some users reporting errors in the log like you can see here: [Websocket connection not active error in logs](https://github.com/briis/unifiprotect/issues/297#issuecomment-917531625)

This was introduced in this PR, but I am not sure what would cause the system to report this error. Can any of you help?

bdraco:
I'm wondering if it was always falling back to polling and not it's being reported.... or it's a change in the new version of protect. 

Will test in the morning to make sure it's not falling back on any of my production installs.

AngellusMortis:
So mine is working right now, but in general I have seen it call back to polling a number of times before (been setting up and testing other unrelated integrations). 

I do not think it is related to anything from here. I think it either a random networking issue or a Home Assistant change. Reason I think it might be HA related is a lot of my other integrations have been getting networking errors on setup as well. Sonos, Unifi Network, the IP printer one etc. I _think_ there is a race condition or something where the integrations are trying to get set before the network is actually fully working. 

When it happens for Unifi Protect, reloading the integration fixes it, though it should be self correcting since it tries to reinitialize the Websocket on every call to `update`. Maybe make the log message a debug/info unless it fails at least 3+ times?

bdraco:
> So mine is working right now, but in general I have seen it call back to polling a number of times before (been setting up and testing other unrelated integrations). 
> 
> 
> 
> I do not think it is related to anything from here. I think it either a random networking issue or a Home Assistant change. Reason I think it might be HA related is a lot of my other integrations have been getting networking errors on setup as well. Sonos, Unifi Network, the IP printer one etc. I _think_ there is a race condition or something where the integrations are trying to get set before the network is actually fully working. Ever since a few versions of HA ago, I have to reload the default Weather integration (Met.no) every time I restart HA for it to work.
> 

Is there an issue open on this?
> 
> 
> When it happens for Unifi Protect, reloading the integration fixes it, though it should be self correcting since it tries to reinitialize the Websocket on every call to `update`. Maybe make the log message a debug/info unless it fails at least 3+ times?
> 
> 
That seems like a reasonable change to me. 
> 
>

AngellusMortis:
> Is there an issue open on this?

I have no idea. I has not annoyed me enough to look into it more. haha. It is basically just extra log spam on startup.

briis:
Thank you both. It confirms what I thought, but i was unsure if there was actually a change done. I will update the issue.

PR #35, opened 2021-08-27
Adds sample data generation for future test writing

AngellusMortis:
**DEPENDS ON**: https://github.com/briis/pyunifiprotect/pull/36

Another PR! This one adds sample data that can later be used for writing tests with.

### Changes

* Adds `jq` to VS Code container
* Adds `generate-sample-data` subcommand to make sample data for tests
    * Data is anonymized by default. Can use `--actual` to disable
    * Pulls all currently used GET inputs and dumps data into JSON
    * Gathers 120 seconds of WS messages and dumps data into JSON
    * Output is `tests/sample_data` by default, use `--output` to change
* Adds new `test_util` submodule used by the `generate-test-data` subcommand to actually generate the data
* Adds `decode-ws-msg` subcommand to help with debugging WS messages
* Adds new `generate-test-data` as a launch configuration in VS Code for debugging
* Adds new `ModelType` enum in `unifi_data` of all of the possible choices of `modelKey`.
* Adds `ws_callback` to `BaseApiClient` so you can add extra raw WS subscribers
* Adds `_get_light_detail` to `UpvServer` so sample light data can be generated.
* Adds VS Code task to update the requirements

### Generating Test Data

The process for generating sample data is

```bash
unifi-protect generate-sample-data
# wait ~120 seconds, maybe trigger some events
# review JSON files in `tests/sample_data` to make sure there is no identifiable data in there.
```


briis:
I am by far, not as experienced with Python as the two of you, and honestly I am not sure how to use all this to run the tests. 
I understand that I can run `unifi-protect COMMAND` in the Dev Container, but what will be the proper series of commands to ensure everything is tested?

Sorry for my ignorance, but I am still learning.

AngellusMortis:
So this is just the first part. Adding the ability to generate data that the tests can then use. Generating for unit test you do not want to make HTTP requests. Especially here. You cannot make pytest call out to a real Unifi Protect server easily because that means putting your credentials somewhere. So we generation all of the data before hand.

Then you can use [unittest.mock.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) to bascially replace the function that actually makes the API requests (in this case `pyunifiprotect.unifi_protect_server.BaseAPIClient.api_request`). 

That will allow writing tests for everything _else_ that uses the data that comes back from Unifi. So we can make sure `_process_ws_message` is working. `server_information`, `_process_cameras_json`, etc. All of the code that is the actual stuff that the library does with the data it gets from Unifi Protect.

That might not sound very useful yet, but it means that in the future, I can edit `_prcess_cameras_json` and make it do more or something else and then be certain that I am not going to break something else because the tests still pass, meaning it still has the same end result.

AngellusMortis:
The other thing it allows is this: https://github.com/AngellusMortis/pyunifiprotect/runs/3450756099?check_suite_focus=true

I just set this up this morning. I set up a self-hosted Github Actions runner that is on my home network. My fork of pyunifiprotect has an action that runs daily at 4 PM UTC (~12 PM ET) that can update the sample data and then once I add some tests, runs the tests against the data. So we can test breaking changes against Unifi Protect in real time, which is pretty neat.

AngellusMortis:
Moved all of the packet refactor code I just made to a separate PR to prevent the bloat on this PR: https://github.com/briis/pyunifiprotect/pull/36

AngellusMortis:
Okay. That should be everything except generating the data for the viewport. I should be getting one in tomorrow and I can add that and be done with everything.

bdraco:
Functional testing passed
Websocket still works

Screen Shot 2021-09-01 at 8 40 27 AM

bdraco (approved):

PR #38, opened 2021-09-02
Adds mypy linting

AngellusMortis:
* Adds mypy and other `types-*` libraries for third packages already used
* Fixes all typing lint errors

### DIFF
--- a/pyunifiprotect/unifi_data.py
+++ b/pyunifiprotect/unifi_data.py
@@ -11,7 +11,7 @@ import json
 import logging
 import struct
 import time
-from typing import Optional, Type
+from typing import Any, Dict, Optional, Type
 import zlib
 
 from .exceptions import WSDecodeError
@@ -39,7 +39,7 @@ PRIVACY_OFF = [[0, 0], [0, 0], [0, 0], [0, 0]]
 PRIVACY_ON = [[0, 0], [1, 0], [1, 1], [0, 1]]
 ZONE_NAME = "hass zone"
 
-PROCESSED_EVENT_EMPTY = {
+PROCESSED_EVENT_EMPTY: Dict[str, Any] = {
     "event_start": None,
     "event_score": 0,
     "event_thumbnail": None,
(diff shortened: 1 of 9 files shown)


pyunifiprotect/unifi_protect_server.py
```diff
@@ -370,7 +372,7 @@ def devices(self):
     async def update(self, force_camera_update=False) -> dict:
         """Updates the status of devices."""
 
-        current_time = time.time()
+        current_time = time.monotonic()
```
bdraco, line 375: Was this intended for this PR?

AngellusMortis: Oh yeah. I forgot to put that in the description. `time.monotonic` is actually what is wanted here. I only noticed it because of fixing the typing around the variables.

`time.time` is actually subject to NTP updates. So you _could_ call `time.time` twice in a row and get a time _before_ the one you got when you previously called it.

`time.monotonic` never changes. It will always "X second python has been running". 

I did double check that were not actually using `_last_device_update_time` and `_last_websocket_check` as actual time markers, but rather just "time since we last did something" so it should be safe to include in this PR.

AngellusMortis: That is a good point. Tracking for change log is good. I will make a separate PR just for that change first so you can merge it separately.

pyunifiprotect/unifi_data.py
```diff
@@ -878,11 +878,11 @@ def packed(self):
 
 
 class WSJSONPacketFrame(WSRawPacketFrame):
-    data: dict = {}
+    data: dict = {}  # type: ignore
```
bdraco, line 881: It would be nice to fully type this at some point in the future

bdraco (approved):

pyunifiprotect/unifi_data.py
AngellusMortis: This type ignore is because the subclass returns raw binary for `data` instead of a `dict`. Typing the dict would likely be pretty hard since it seems to change a lot based on what was updated (partials).

My plan was to make more typed classes like this for Cameras, Lights, etc. so we can type them better. So the further you get away from the raw Protect data, the better it is typed. By the time to gets to HA it should be 100% typed. But that is still a couple of PRs away.

PR #40, opened 2021-09-04
Fixes inconsistencies with setting up websocket

AngellusMortis:
Fixes a race condition / random bug introduced by changing from `time.time` to `time.monotonic` by actually getting the diff between the last check markers and the current monotonic clock.

The issue was since the monotonic clock is not real time, it can start at 0 and be a number less then the check intervals. So if `current_time = 30`, the `(current_time - WEBSOCKET_CHECK_INTERVAL_SECONDS) > self._last_websocket_check` would fail and the Websocket would never be connected.

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -345,8 +345,8 @@ class UpvServer(BaseApiClient):  # pylint: disable=too-many-public-methods, too-
     _event_state_machine: ProtectEventStateMachine = ProtectEventStateMachine()
     _device_state_machine: ProtectDeviceStateMachine = ProtectDeviceStateMachine()
     _processed_data: Dict[str, dict] = {}
-    _last_websocket_check: float = 0
-    _last_device_update_time: float = 0
+    _last_websocket_check: float = -WEBSOCKET_CHECK_INTERVAL_SECONDS
+    _last_device_update_time: float = -DEVICE_UPDATE_INTERVAL_SECONDS
     _ws_subscriptions: List[Callable[[Dict[str, dict]], None]] = []
     _is_first_update: bool = True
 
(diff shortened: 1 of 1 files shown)


pyunifiprotect/unifi_protect_server.py
```diff
@@ -345,8 +345,8 @@ class UpvServer(BaseApiClient):  # pylint: disable=too-many-public-methods, too-
     _event_state_machine: ProtectEventStateMachine = ProtectEventStateMachine()
     _device_state_machine: ProtectDeviceStateMachine = ProtectDeviceStateMachine()
     _processed_data: Dict[str, dict] = {}
-    _last_websocket_check: float = 0
-    _last_device_update_time: float = 0
+    _last_websocket_check: float = -WEBSOCKET_CHECK_INTERVAL_SECONDS
```
bdraco, line 348: Can we use None here to indicate it's never been check?

Or at least something like NEVER_TIME = -100000 for both of these?

AngellusMortis: `None` would make the typing odd. A super negative number would work as well, but setting them to their inverse of what they are compared against essentially does the same thing.

bdraco: It's likely not going to be obvious a few years from now to someone looking at the code.

Ideally the code should show intent.
I like the negative constant because it's more clear (at least to me) that we are matching against is a value that is never expected to come from the data source.

If you prefer to keep the existing construct, please add comment about what is going on with the negative number so we remember later.

Thanks

AngellusMortis: The constant does make sense for future looking. That is as good of reason as any to use it.

bdraco (approved):

bdraco (approved):
Testing Looks good. 👍 
Thanks!

Screen Shot 2021-09-04 at 10 34 59 AM

PR #45, opened 2021-09-11
Adds more tests

AngellusMortis:
So the plan going forward is write more tests for areas of code I want to refactor / add better typing to and then change that code. I can try to include them in single PR or make two separate ones. In this case I am doing two separate ones since this PR was starting to get big. 

Stuff this PR does:

* Fixes a number of places in the sample data generation where anonymized IDs were not lining up perfectly
* adds tests to test all of the "get" endpoints in `UpvServer`
* adds tests for all of the `process_<type>` functions (except `process_sensor`, but that is because I cannot get me handles on Smart Sensor since it is EA and sold out)
    * These tests might seem kind of silly since the `legacy_process_*` methods are basically just refactored versions of the existing ones, but the idea is those are the parity check with the existing ones without changing them. The plan for the next PR is to gut and refactor the existing ones to use real typed Python objects (likely via pydantic). 
* Adds more data enums for events / states.

These tests have been running once a day for over a week over on my fork against my live Unifi data using a self-hosted GHA runner:

https://github.com/AngellusMortis/pyunifiprotect/actions/workflows/update-data.yml


bdraco (approved):
LGTM 👍

bdraco:
Testing now...

bdraco:
Manual testing looks good

PR #47, opened 2021-09-20
Adds more docs for setup/development

AngellusMortis:


briis:
Thanks for this. Just tried it all out, and it works as described.

PR #49, opened 2021-09-24
Improves determining reason for bad responses

AngellusMortis:
* Adds new method to try to retrieve error from response body when there is a bad error code instead of always using `response.reason`, which will always be the "name" of the HTTP status code (500 = Internal Server Error, 400 = Bad Request, etc.)
* Handles error response in `write_image_file` for generating sample data

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -68,6 +68,7 @@ class BaseApiClient:
     _verify_ssl: bool
     _is_authenticated: bool = False
     _is_unifi_os: Optional[bool] = None
+    _websocket_failures: int = 0
 
     req: aiohttp.ClientSession
     headers: Optional[dict] = None
@@ -336,6 +337,16 @@ class BaseApiClient:
             _LOGGER.debug("websocket disconnected")
             self.ws_connection = None
 
+    def _log_websocket_failure(self):
+        self._websocket_failures += 1
+        log = _LOGGER.warning
+        # default polling interval is every 2 seconds
+        # wait until 1 minute before reporting the errors
+        # as warnings
+        if self._websocket_failures < 30:
+            log = _LOGGER.info
+        log("Unifi OS: Websocket connection not active, failing back to polling")
+
     def _process_ws_message(self, msg):
         raise NotImplementedError()
 
@@ -400,7 +411,7 @@ class UpvServer(BaseApiClient):  # pylint: disable=too-many-public-methods, too-
             return self._processed_data if device_update else {}
 
         if self.is_unifi_os:
-            _LOGGER.warning("Unifi OS: Websocket connection not active, failing back to polling")
+            self._log_websocket_failure()
 
         self._reset_device_events()
         updates = await self._get_events(lookback=10)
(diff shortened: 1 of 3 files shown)


briis:
I am getting a few bugs reported on this:

> [33m2021-09-23 14:00:45 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling

Will this fix this issue?

AngellusMortis:
No, it will only help if there is an expected HTTP failure. 

Instead of:
> Request failed: /proxy/protect/api/cameras/6101cf1100acde0387003c11/snapshot - Status: 500 - Reason: Internal Server Error

You will see:
> Request failed: /proxy/protect/api/cameras/6101cf1100acde0387003c11/snapshot - Status: 500 - Reason: Unable to get snapshot for Garage Camera [E063DA0041D4 @ 192.168.51.24], camera not accessible.

Which is a much better error message. Based on the issue on the other project, it looks like it would help some users with debugging the issue on their end if they are getting the generic 500 error.

Based on our previous discussions, it sounds like the falling back to polling could just be a startup error that _should_ go away. Though some users are reporting it not going away. I just added another comment to add the logging improvements @bdraco and discussed before. The default polling interval is every 2 seconds, so I changed it to logging the fallback message as a `info` instead of a `warning` for the first 30 (60 seconds) of attempts.

briis:
Perfect Christopher. I'll ask Nick to review.

pyunifiprotect/utils.py
```diff
@@ -1,11 +1,28 @@
...
+    reason = str(response.reason)
+
+    try:
+        json = await response.json()
+        reason = json.get("error", str(json))
+    except Exception:  # pylint: disable=broad-except
+        try:
+            reason = await response.text()
+        except Exception:  # pylint: disable=broad-except
+            pass
```
bdraco, line 21:
```suggestion
        with contextlib.suppress(Exception):  # pylint: disable=broad-except
            reason = await response.text()
```

AngellusMortis: Neat, I did not know that existed.

bdraco:
Looks good. 👍  Will do some testing after lunch

pyunifiprotect/unifi_protect_server.py
```diff
@@ -329,11 +338,24 @@ async def _setup_websocket(self):
...
+    def _log_websocket_failure(self):
+        now = datetime.now()
+        if self._websocket_error_start is None:
+            self._websocket_error_start = now
+
+        self._websocket_failures += 1
+        log = _LOGGER.warning
+
+        if self._websocket_failures < 10 or (now - self._websocket_error_start) < WEBSOCKET_ERROR_GRACE_PERIOD:
+            log = _LOGGER.info
```
bdraco, line 356:
```suggestion
            log = _LOGGER.debug
```

A lot of users have their logging set at info and will still generate issues from this.  I think we should leave this at debug

bdraco (approved):
Tested, LGTM 👍

briis:
With the latest PR's submitted and approved, is this ready for deployment? Then I wanted to make a new release to HA using this new version.

AngellusMortis:
The other PR should not really affect anything with the existing `unifiprotect` integration since I isolated the new stuff to a different API Class. Cutting a new release of the integration now should be a great idea.

briis:
Perfect. I will build this tomorrow when I am back, and make a new release of the integration using this build within the next days.

PR #50, opened 2021-09-25
Fixes tests if test camera is offline

AngellusMortis:
Test failure found from latest test run against live data: https://github.com/AngellusMortis/pyunifiprotect/runs/3708943132


AngellusMortis:
Any objections to merging this one? It is test only.

@bdraco / @briis

briis:
No, go ahead.

Issue #33, opened 2021-08-07
Issue with CSRF token


ps1337:
hey there and thanks for this awesome library. I've encountered an issue in my particular setup:

- I'm running Unifi Protect on Unifi OS
- For Unifi OS machines, this library assumes that server responses include a `x-csrf-token`
- There's a special case where this causes issues: If requests originate from localhost, Unifi OS will not include this header and does not require this header to be present.
- If this header is missing in server responses, subsequent requests cause exceptions because this value is `None` on our side, e.g.:

```
  File "aiohttp/_http_writer.pyx", line 137, in aiohttp._http_writer._serialize_headers
  File "aiohttp/_http_writer.pyx", line 109, in aiohttp._http_writer.to_str
TypeError: Cannot serialize non-str key None
```

Therefore, I suggest adding a check in `check_unifi_os()` for this special case where `localhost` or `127.0.0.1` is present in `self_base_url`. The `authenticate()` method should then also not try to include this header.

Let me know :)

Thanks!

briis:
I am happy with the solution, but could you explain the use case where you need to use `localhost`? Are you running non UniFi software on the UniFi HW?

ps1337:
The other way around actually, I'm running unifi protect on my raspi 4 and I'm using home assistant on the same machine, that's why my use case involves using localhost

-----Original Message-----

briis:
Got it. I will look in to this, but I will need you to test it, as I don't have that setup.

ps1337:
Thanks! Sure thing, I've already patched it locally on my side (in a hacky way) so just ping me 

-----Original Message-----

briis:
Please post here what you did and I can incorporate that.

ps1337:
Sure, here's the whole file so you can `diff` it:

```python
"""Unifi Protect Server Wrapper."""

import asyncio
import datetime
import json as pjson
import logging
import time
from typing import Optional

import aiohttp
from aiohttp import client_exceptions
import jwt

from .const import SERVER_ID, SERVER_NAME
from .unifi_data import (
    DEVICE_MODEL_LIGHT,
    EVENT_MOTION,
    EVENT_RING,
    EVENT_SMART_DETECT_ZONE,
    PRIVACY_OFF,
    PRIVACY_ON,
    PROCESSED_EVENT_EMPTY,
    TYPE_RECORD_NEVER,
    ZONE_NAME,
    ProtectDeviceStateMachine,
    ProtectEventStateMachine,
[...]

PR #67, opened 2021-10-30
Adds support for no CSRF token from UnifiOS

AngellusMortis:
If you are using `pyunifiprotect` _on_ the same physical machine as your UnifiOS console, UnifiOS does not provide a CSRF token. `aiohttp` cannot handle a header with a `NoneType` so this should exclude the CSRF token if it is not being provided by UnifiOS.

Fixes https://github.com/briis/pyunifiprotect/issues/33 and https://github.com/briis/unifiprotect/issues/312

## HA Install Link:

```
git+https://github.com/AngellusMortis/pyunifiprotect.git@feature/no-csrf-token#pyunifiprotect==0.35.3
```

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -207,9 +207,13 @@ class BaseApiClient:
 
         response = await self.request("post", url=url, json=auth)
         self.headers = {
-            "x-csrf-token": response.headers.get("x-csrf-token"),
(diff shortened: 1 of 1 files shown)


bdraco (approved):
LGTM.

edwardmp:
Can confirm this solves the issue.

bdraco:
I'll pull this into my test branch and report back

PR #69, opened 2021-10-30
Enforces strict mypy type checking

AngellusMortis:
Last dozy of a PR, I swear. 

Enables a lot more strict typing flags for mypy and adds the needed type hints everywhere.

The major changes outside of type hints:

* Adds defaults to various places we use `.get` on a dict
* Removes `raw` from `api_request` and splits that one call into four to help with typing: `api_request_raw` (returns `bytes`), `api_request` (returns `dict` | `list`), `api_request_obj` (returns `dict`) and `api_request_list` (returns `list`)

I can move these changes into smaller PRs, but for this, it might be better to just test everything all at once.

## HA Install Link:

```
git+https://github.com/AngellusMortis/pyunifiprotect.git@feature/strict-mypy#pyunifiprotect==0.35.3
```

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -614,7 +614,7 @@ class SensorStats(ProtectBaseObject):
     humidity: SensorStat
     temperature: SensorStat
 
-    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {
+    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {  # type: ignore
         "light": SensorStat,
         "humidity": SensorStat,
         "temperature": SensorStat,
@@ -642,7 +642,7 @@ class Sensor(ProtectAdoptableDeviceModel):
     # TODO:
     # mountType
 
-    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {
+    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {  # type: ignore
         "alarmSettings": SensorSettingsBase,
         "batteryStatus": SensorBatteryStatus,
     }
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -70,7 +70,7 @@ class Light(ProtectMotionDeviceModel):
     is_camera_paired: bool
 
     UNIFI_REMAP: ClassVar[Dict[str, str]] = {**ProtectMotionDeviceModel.UNIFI_REMAP, **{"camera": "cameraId"}}
-    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {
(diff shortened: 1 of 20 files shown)


bdraco:
Looks like it conflicts with the one I just merged

AngellusMortis:
Yep, I expected that. I fill fix now.

AngellusMortis:
Type ignores:

* `PROTECT_OBJ_FIELDS`: it should be `Dict[str, Type[ProtectBaseObject]]`, but that causes all kinds of issues with Pydantic. This was best way I could get it to work (other suggestions very welcome)
* `write_json_file`: return type is equal to input type. Not sure how to model that with typing. A separate method would be a PITA since it is just test code generation.

AngellusMortis:
I figured out `write_json_file`! I can solve that one with the `overload` decorator from mypy.

pyunifiprotect/unifi_protect_server.py
```diff
@@ -5,13 +5,12 @@
 import json as pjson
```
bdraco, line 5: This one should be safe to add `from __future__ import annotations` since we do it almost everywhere in core already

AngellusMortis: Probably a good call. I have only been doing it in places that I have actually needed it (self-referencing annotations).

pyunifiprotect/data/devices.py
```diff
@@ -70,7 +70,7 @@ class Light(ProtectMotionDeviceModel):
     is_camera_paired: bool
 
     UNIFI_REMAP: ClassVar[Dict[str, str]] = {**ProtectMotionDeviceModel.UNIFI_REMAP, **{"camera": "cameraId"}}
-    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {
+    PROTECT_OBJ_FIELDS: ClassVar[Dict[str, Callable]] = {  # type: ignore
```
bdraco, line 73: Ideally we get rid of the ignores in the future

AngellusMortis: Yep, I actually solved that in the other PR (mostly) if you check that one. Both of them will very much conflict each other, so which ever one gets merged first, the other one will need clean up.

bdraco: I'm already about half way through this one and have it installed for testing so this one first

AngellusMortis: That is actually preferred as well since this one actually needs testing. Get the strict enforcement going will help all future code changes as well.

bdraco:
Testing looks good

- [x] Snapshots
- [x] Streaming
- [x] Websocket
- [x] api calls
- [x] no visible errors in the log
Screen Shot 2021-11-02 at 10 29 33 AM
Screen Shot 2021-11-02 at 10 29 25 AM
Screen Shot 2021-11-01 at 11 30 29 AM

bdraco (approved):
This looks great 👍

PR #70, opened 2021-10-31
Makes nested Protect Objects smarter

AngellusMortis:
When trying to make everything be able to use construct, I kept running into issues with how I was handled child `ProtectBaseObject` attributes. This is a big refactor to start handling them all automatically instead of trying to manually manage them, which was a nightmare. 

This also supports attributes that have many child `ProtectBaseObject` values (lists / dicts) which there was no real support for before.

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -34,19 +34,37 @@ T = TypeVar("T", bound="ProtectBaseObject")
 
 
 class ProtectBaseObject(BaseModel):
+    """
+    Base class for building Python objects from Unifi Protect JSON.
+
+    * Provides `.clean_unifi_dict` to convert UFP JSON to a more Pythonic formatted dict (camel case to snake case)
+    * Add attrs with matching Pyhonic name and they will automatically be populated from the UFP JSON if passed in to the constructer
+    * Provides `.unifi_dict` to convert object back into UFP JSON
+    """
+
     _api: Optional[ProtectApiClient] = PrivateAttr(None)
     _initial_data: Dict[str, Any] = PrivateAttr()
 
     _protect_objs: ClassVar[Optional[Dict[str, Callable]]] = None
     _protect_lists: ClassVar[Optional[Dict[str, Callable]]] = None
     _protect_dicts: ClassVar[Optional[Dict[str, Callable]]] = None
-
-    UNIFI_REMAP: ClassVar[Dict[str, str]] = {}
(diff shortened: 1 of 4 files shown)


AngellusMortis:
@bdraco Feel free to review if you want/have time. Thankfully now there should not need to be a big HA testing cycle so it should save you a bunch of time.

pyunifiprotect/data/base.py
```diff
@@ -32,64 +33,254 @@
...
         data = self.clean_unifi_dict(data)
+        del data["api"]
         super().__init__(**data)
 
         self._api = api
 
+    @classmethod
+    def _get_unifi_remaps(cls) -> Dict[str, str]:
+        """
+        Helper method for overriding in child classes for remapping UFP JSON keys to Python ones that do not fix the
```
bdraco, line 77:
```suggestion
        Helper method for overriding in child classes for remapping UFP JSON keys to Python ones that do not fit the
```

pyunifiprotect/data/base.py
```diff
@@ -32,64 +33,254 @@
...
+    def _clean_protect_obj_dict(
+        cls, items: Dict[Any, Any], klass: Type[ProtectBaseObject], api: Optional[ProtectApiClient]
+    ) -> Dict[Any, Any]:
+        cleaned_items: Dict[Any, Any] = {}
+        for key, value in items.items():
+            cleaned_items[key] = cls._clean_protect_obj(value, klass, api)
+        return cleaned_items
+
     @classmethod
     def clean_unifi_dict(cls, data: Dict[str, Any]) -> Dict[str, Any]:
```
bdraco, line 172: Would it make sense to name this `convert_ufp_to_native` or something similar?

bdraco (changes requested):
Looks good.  On question on the naming.

Marking this request changes so it pings back on re-review request

pyunifiprotect/data/base.py
AngellusMortis: That naming could be better. The `clean_` notation came from me doing Django so much. `clean_` is just basically the common format in Django for validating/"cleaning" data. 

But since we are not in Django, `unifi_dict_to_dict` might make more sense? Since `.dict()` returns a "Python dict" and `.unifi_dict()` returns a "UFP dict"

bdraco: > unifi_dict_to_dict

WFM 👍

bdraco (approved):

PR #72, opened 2021-11-01
Makes test_{device} tests optional

AngellusMortis:


AngellusMortis:
Fix for new test failure from my workflow to test against live data. Will merge after tests pass.

PR #74, opened 2021-11-02
Improves reconnect/logging logic for Websockets

AngellusMortis:
Related: https://github.com/briis/unifiprotect/issues/321

Because we currently log every time update is called that the WS disconnected after it has been disconnected for 60 seconds or more, there is no actual way to know if the log messages were valid or not. `.update` will only try to reconnect if there has been at least 120 seconds last attempt. So the logs could just be false positives.

This should change it so the warning log message only when a reconnect attempt is made. This should greatly reduce the log spam and only tell us when real reconnect attempts are made to help us debug not reconnecting better.

I also cleaned up and made both `.update` methods work basically the same as each other (I am the one that recommended using `monotonic` and then went and used datetimes in the new one, oops). 

## HA Install Link:

```
git+https://github.com/AngellusMortis/pyunifiprotect.git@bugfix/ws-reconnect#pyunifiprotect==0.35.7
```

### DIFF
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -374,10 +374,10 @@ class BaseApiClient:
             await self.async_connect_ws()
 
         # log if no active WS
-        if not self.ws_connection:
+        if not self.ws_connection and not first_check:
             log = _LOGGER.debug
             # but only warn if a reconnect attempt was made
-            if not first_check and connect_ws:
+            if connect_ws:
                 log = _LOGGER.warning
             log("Unifi OS: Websocket connection not active, failing back to polling")
 
--- a/pyunifiprotect/unifi_protect_server.py
+++ b/pyunifiprotect/unifi_protect_server.py
@@ -1,4 +1,5 @@
 """Unifi Protect Server Wrapper."""
+from __future__ import annotations
 
 import asyncio
 from datetime import datetime, timedelta
@@ -53,17 +54,16 @@ from pyunifiprotect.unifi_data import (
 from pyunifiprotect.utils import get_response_reason, to_js_time
 
 NEVER_RAN = -1000
-DEVICE_UPDATE_INTERVAL_SECONDS = 60
-DEVICE_UPDATE_INTERVAL = timedelta(seconds=DEVICE_UPDATE_INTERVAL_SECONDS)
-WEBSOCKET_CHECK_INTERVAL_SECONDS = 120
-WEBSOCKET_CHECK_INTERVAL = timedelta(seconds=WEBSOCKET_CHECK_INTERVAL_SECONDS)
+# how many seconds before the bootstrap is refreshed from Protect
+DEVICE_UPDATE_INTERVAL = 60
+# how many seconds before before we check for an active WS connection
+WEBSOCKET_CHECK_INTERVAL = 120
 LIGHT_MODES = ["off", "motion", "always"]
 LIGHT_ENABLED = ["dark", "fulltime"]
 LIGHT_DURATIONS = [15000, 30000, 60000, 300000, 900000]
 
 DEFAULT_SNAPSHOT_WIDTH = 1920
 DEFAULT_SNAPSHOT_HEIGHT = 1080
-WEBSOCKET_ERROR_GRACE_PERIOD = timedelta(seconds=60)
 
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bdraco:
This looks good.  Considering how much this has come up, how about a test for this?

tests/test_unifi_protect_server.py
```diff
@@ -118,3 +122,50 @@ async def test_get_events(protect_client: ProtectApiClient, raw_events):
...
+
+    assert active_ws is False
+
+    expected_logs = ["Unifi OS: Websocket connection not active, failing back to polling"]
+    assert expected_logs == [rec.message for rec in caplog.records]
+    assert caplog.records[0].levelname == "DEBUG"
+
+
+@pytest.mark.asyncio
+async def test_check_ws_reconnect(protect_client: ProtectApiClient, caplog):
```
bdraco, line 159:
```suggestion
async def test_check_ws_reconnect(protect_client: ProtectApiClient, caplog: pytest.LogCaptureFixture):
```

tests/test_unifi_protect_server.py
```diff
@@ -118,3 +122,50 @@ async def test_get_events(protect_client: ProtectApiClient, raw_events):
...
+    protect_client.ws_connection = None
+
+    active_ws = await protect_client.check_ws()
+
+    assert active_ws is True
+    assert ["Checking websocket"] == [rec.message for rec in caplog.records]
+
+
+@pytest.mark.asyncio
+async def test_check_ws_no_ws(protect_client: ProtectApiClient, caplog):
```
bdraco, line 142:
```suggestion
async def test_check_ws_no_ws(protect_client: ProtectApiClient, caplog: pytest.LogCaptureFixture):
```

tests/test_unifi_protect_server.py
```diff
@@ -118,3 +122,50 @@ async def test_get_events(protect_client: ProtectApiClient, raw_events):
     assert len(events) == len(expected_events)
     for index, event in enumerate(events):
         compare_objs(event.model.value, expected_events[index], event.unifi_dict())
+
+
+@pytest.mark.asyncio
+async def test_check_ws_initial(protect_client: ProtectApiClient, caplog):
```
bdraco, line 128:
```suggestion
async def test_check_ws_initial(protect_client: ProtectApiClient, caplog: pytest.LogCaptureFixture):
```

bdraco:
IMG_0134
IMG_0135


Motion detection working as expected

bdraco:
Screen Shot 2021-11-02 at 11 50 40 AM

AngellusMortis:
I have also been running it since I added the tests. No log messages yet. I am turning off my Protect to test the logs and reconnecting.

pyunifiprotect/unifi_protect_server.py
```diff
@@ -417,31 +423,30 @@ def devices(self) -> Dict[str, Dict[str, Any]]:
     async def update(self, force_camera_update: bool = False) -> Dict[str, Any]:
         """Updates the status of devices."""
 
-        current_time = time.monotonic()
-        device_update = False
+        now = time.monotonic()
+        if force_camera_update and self.ws_connection is not None:
```
bdraco, line 427:
```
    async def async_turn_on(self, **kwargs):
        """Turn the device on."""
        if self._switch_type == "hdr_mode":
            _LOGGER.debug("Turning on HDR mode")
            await self.upv.set_camera_hdr_mode(self._device_id, True)
        elif self._switch_type == "high_fps":
            _LOGGER.debug("Turning on High FPS mode")
            await self.upv.set_camera_video_mode_highfps(self._device_id, True)
        else:
            _LOGGER.debug("Changing Status Light to On")
            await self.upv.set_device_status_light(
                self._device_id, True, self._device_type
            )
        await self.protect_data.async_refresh(force_camera_update=True)
```

I think this is going to be an issue since In `switch.py` we have this.

AngellusMortis: I think that would still be expected behavior. When we update the devices directly from the boostrap, we would want to reset the Websocket as well. If we do not, it can create a weird race condition where a WS message comes in and updates the devices after we pull in the new bootstrap but be from _before_ the bootstrap was generated.

bdraco: Won't we miss motion events every time a switch is flipped due to the connection drop/reconnect?

AngellusMortis: Fair point. Probably is best to leave it as is until we can update the HA side to stop it from doing a force because the inverse is also a potentional issue since async is still single threaded:

1. You call update
2. A new WS message comes in (but is not processed, single thread)
3. You fetch a new bootstrap
4. Devices are reset
5. WS message is applied. Changing the state of the devices

bdraco (approved):

bdraco (approved):

PR #75, opened 2021-11-03
Updates ProtectBaseObject to use construct instead of normal pydantic constructor

AngellusMortis:
The main goal of this is to remove Pydantic's hard runtime type enforcement when creating objects from JSON that comes from UFP. Runtime type enforcement will still be applied to manual updates to UFP objects. To enable type enforcement for UFP JSON, set the ENV `UFP_DEBUG=True`

* Removes `klass_from_dict` and `from_unifi_dict` out of `ProtectBaseObject` and made them separate methods.
* Changed `__init__` to not run `unifi_dict_to_dict`
* Adds `from_unifi_dict` that runs `unifi_dict_to_dict` and then chooses between `__init__` and `construct` based on ENV variable
* Adds support for construct generating the same object structure as `__init__`
* Adds `convert_unifi_data` to take a pydantic field and automatically convert the passed in value to the correct type (usually just passing the raw value into the class constructor)
* Fixes comparison of pydantic's Color objects
* Adds new tests to benchmark `__init__` vs. `construct`.
   * `construct` is _slightly_ faster, but not a significant amount. They are both about the same performance. 
* Forces all color output for pytest for GHA.

## HA Install Link:

```
git+https://github.com/AngellusMortis/pyunifiprotect.git@feature/pydantic-construct#pyunifiprotect==0.35.7
```

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -21,7 +21,13 @@ from pydantic.fields import SHAPE_DICT, SHAPE_LIST, PrivateAttr
 
 from pyunifiprotect.data.types import ModelType, StateType
 from pyunifiprotect.exceptions import BadRequest, DataDecodeError
-from pyunifiprotect.utils import process_datetime, serialize_unifi_obj, to_snake_case
+from pyunifiprotect.utils import (
+    convert_unifi_data,
+    is_debug,
+    process_datetime,
+    serialize_unifi_obj,
+    to_snake_case,
+)
 
 if TYPE_CHECKING:
     from pyunifiprotect.data.devices import Bridge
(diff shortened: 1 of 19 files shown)


pyunifiprotect/data/base.py
```diff
@@ -21,15 +21,99 @@
...
+    )
+
+    if "modelKey" not in data:
+        raise DataDecodeError("No modelKey")
+
+    model = ModelType(data["modelKey"])
+
+    klass: Optional[Type[ProtectModel]] = None
+
+    if model == ModelType.EVENT:
```
bdraco, line 71: Is it possible to avoid the linear search for the model ?

pyunifiprotect/data/base.py
```diff
@@ -21,15 +21,99 @@
...
+ProtectObject = TypeVar("ProtectObject", bound="ProtectBaseObject")
+
+
+def get_klass_from_dict(data: Dict[str, Any]) -> Type[ProtectModel]:
+    """
+    Helper method to read the `modelKey` from a UFP JSON dict and get the correct Python class for conversion.
+    Will raise `DataDecodeError` if the `modelKey` is for an unknown object.
+    """
+
+    from pyunifiprotect.data.devices import (  # pylint: disable=import-outside-toplevel
```
bdraco, line 47: What's the reasoning for the late imports?

AngellusMortis: Recursive imports. I tried to completely get rid of them at one point, but `Bootstrap` needs this method to be able to process events. I could use some importlib magic to import them on the fly and that would remove the imports here.

pyunifiprotect/data/base.py
AngellusMortis: If I moved the classes to the `.data.types` module (to make it a hashmap), it would likely cause even more of a mess. I could try to just use the class name in `.data.types` and then use importlib to dynamically import the class (would remove the imports below as well)

AngellusMortis: I think refactoring this here it out of scope for this PR (I only moved the code), but if you have a suggestion for improving the search and the imports, I will definitely make a follow up PR to improve them.

bdraco (approved):

pyunifiprotect/data/base.py
AngellusMortis: Oh cool. I really look forward to seeing what you come up with.

PR #76, opened 2021-11-03
Adds PEP 561 typing metadata

AngellusMortis:
* Adds metadata needed so type hints are exported for the package when installed.
* Fixes args for `ProtectApiClient` so they appear in type hints correctly.


bdraco (approved):
LGTM

PR #80, opened 2021-11-03
Updates from migrating to UNVR-Pro

AngellusMortis:
Migrated from UDM-Pro to UNVR-Pro. Some minor updates to support my new instance. Looks like most of them are just some unexpected defaults when there are no values (storage capacity/last login/etc).


AngellusMortis:
Waiting for my GHA run against my new instance to finish running before merging: https://github.com/AngellusMortis/pyunifiprotect/actions/runs/1419069693

PR #82, opened 2021-11-05
Refactors our new API code into separate module

AngellusMortis:
Refactors `BaseApiClient` and `ProtectApiClient` into a separate module to make the files smaller and separate new API code from existing HA used code.

`test_unifi_protect_server.py` was renamed to `test_api.py` and `test_unifi_protect_server_old.py` was renamed to `test_unifi_protect_server.py`. Thus the diff. No code was actually changed.


bdraco (approved):
Testing looks good

PR #83, opened 2021-11-05
Adds WS Subscriptions to new API Client

AngellusMortis:
The new subscriptions uses a new message class:

```python
class WSSubscriptionMessage:
    action: WSAction
    new_update_id: UUID
    changed_data: Dict[str, Any]
    new_obj: ProtectModel
    old_obj: Optional[ProtectModel] = None
```

The fields are pretty straight forward:

* `old_obj` is `None` if `action` is `ADD`
* `old_obj` + `changed_data` = `new_obj` 

I am not a big fan of `changed_data` being a raw dict, but it is the already posted "cleaned" data so the keys match up with the attrs perfectly.

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -420,7 +420,10 @@ class ProtectApiClient(BaseApiClient):
             return
 
         for sub in self._ws_subscriptions:
-            sub(processed_message)
+            try:
+                sub(processed_message)
+            except Exception:  # pylint: disable=broad-except
+                _LOGGER.exception("Exception while running subscription handler")
 
     @property
     def bootstrap(self) -> Bootstrap:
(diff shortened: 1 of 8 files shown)


pyunifiprotect/api.py
```diff
@@ -412,7 +414,13 @@ async def update(self, force: bool = False) -> Union[Bootstrap, List[Event]]:
...
     def _process_ws_message(self, msg: aiohttp.WSMessage) -> None:
         packet = WSPacket(msg.data)
-        self.bootstrap.process_ws_packet(packet)
+        processed_message = self.bootstrap.process_ws_packet(packet)
+
+        if processed_message is None:
+            return
+
+        for sub in self._ws_subscriptions:
+            sub(processed_message)
```
bdraco, line 423: Should this guard against exceptions in the callback?

AngellusMortis: Probably a good idea.

bdraco (approved):
LGTM 👍

briis:
> LGTM 👍

I have to ask, what does this stand for - me feeling stupid...

bdraco:
Looks good to me 

or sometimes: Lets get this merged

briis:
I need to get in to the Lingo. Thanks

PR #84, opened 2021-11-05
Adds various get_ methods to new API client

AngellusMortis:
*  `BaseApiClient`: Made `api_path` / `ws_path` members instead of properties since they are no longer dynamic
*  `BaseApiClient`: Converted the `_base_url` and `_base_ws_url` into properties so they can be dynamically generated from the `_host` / `_port` values. Also, strips out the default 443 port
* `ProtectApiClient`: Adds `connection_host` property that is an IP from the `nvr.hosts`, preferring the `_host` used by the client
* `ProtectApiClient`: Adds `get_devices` / `get_device` and all modelKey type specific versions
* `ProtectApiClient`: Adds `get_camera_snapshot` for cameras
* `ProtectApiClient`: Adds `get_event_thumbnail` / `get_event_heatmap` for events
* Converts sample data generation to use `ProtectApiClient` instead of `UpvServer`
    * Adds a bunch more sample data
    * Regenerates sample data
* `CameraChannel`: Adds `rtsp_url` / `rtsps_url`
* `Camera`: Adds `timelapse_url` and `get_snapshot`
* `Event`: Adds `get_thumbnail` and `get_snapshot` 
* adoptable Protect devices, `Liveview` and `NVR`: Adds `protect_url`


bdraco:
+8728, -8142 😮

bdraco:
Thankfully mostly testing code changes 👍

pyunifiprotect/utils.py
```diff
@@ -219,3 +220,13 @@ def serialize_list(items: Iterable[Any]) -> List[Any]:
...
     return Decimal(str(round(num, digits)))
+
+
+def ip_from_host(host: str) -> IPv4Address:
+    try:
+        return IPv4Address(host)
+    except AddressValueError:
+        pass
+
+    _, _, ips = socket.gethostbyname_ex(host)
```
bdraco, line 231: Can this fail with `socket.gaierror`.  Should we let it blow up?

AngellusMortis: I am thinking we actually should let that blow up. If the DNS name cannot resolve, the code will have other issues working.

AngellusMortis: Let me do some testing with bad data/bad network and see what the exception looks like. If it is confusing, we can catch it and make it easier to understand.

AngellusMortis: This is what I get from various errors:

```python
>>> socket.gethostbyname("google.com")  # timeout
  File "<stdin>", line 1, in <module>
socket.gaierror: [Errno -3] Temporary failure in name resolution

>>> socket.gethostbyname("notarealdomain.com") # success
'64.90.40.98'

>>> socket.gethostbyname("notarealdomainsdf.com")  # bad DNS
  File "<stdin>", line 1, in <module>
socket.gaierror: [Errno -2] Name or service not known
```

So these are pretty common errors I have seen all the time. I think this one is safe to let it go through. Should be errors people recognize.

AngellusMortis:
Yep, 90% of it is the regeneration of the sample data.

pyunifiprotect/test_util/__init__.py
```diff
@@ -151,120 +150,170 @@ def write_image_file(self, name: str, raw: Optional[bytes]) -> None:
...
+
+        return motion_event
+
+    async def generate_device_data(self, motion_event: Optional[Dict[str, Any]]) -> None:
+        await self.generate_camera_data(motion_event)
+        await self.generate_light_data()
+        await self.generate_viewport_data()
+        await self.generate_sensor_data()
+        await self.generate_bridge_data()
+        await self.generate_liveview_data()
```
bdraco, line 178: Should we do `asyncio.gather` here?

bdraco (approved):
I did some basic testing and didn't find any collateral breakage.

Should be good to go.

One question above.

pyunifiprotect/test_util/__init__.py
AngellusMortis: Great call since I really simplified that method.

I will add gather in a bit.

bdraco (approved):

PR #86, opened 2021-11-06
Adds get_camera_video to API client and more tests

AngellusMortis:
Here starts the decent into ffmpeg madness.


briis:
Looks interesting. What are you trying to achieve with the ffmpeg module? Recode the rtsp stream? (I am very unfamiliar with video streaming)

AngellusMortis:
No, in this case, ffmpeg is only for generating a fake video for tests and then validating the format/length of a video inside of tests. 

There does not appear to be any python only solutions for easily validating/creating video files like there is with images (Pillow).

briis:
Got it, thanks.

pyunifiprotect/api.py
```diff
@@ -723,6 +724,37 @@ async def get_camera_snapshot(
...
+                raise BadRequest from e
+
+        params = {
+            "camera": camera.id,
+            "channel": channel_index,
+            "start": to_js_time(start),
+            "end": to_js_time(end),
+        }
+
+        return await self.api_request_raw("video/export", params=params, raise_exception=False)
```
AngellusMortis, line 756: This here is basically the "meat" of this PR. It just calls to `/proxy/protect/api/video/export` to make Protect create the MP4 clip for us. This what what gets called when you "export" a video in the UI.

tests/conftest.py
```diff
@@ -146,39 +168,60 @@ async def old_protect_client():
...
 
+    await client.async_disconnect_ws()
+    await client.req.close()
+
+    if client.ws_task is not None:
+        client.ws_task.cancel()
+
+    # empty out websockets
+    while client.ws_connection is not None and client.ws_task is not None and not client.ws_task.done():
+        await asyncio.sleep(0.1)
```
bdraco, line 216: It would be cleaner to use asyncio synchronization primitives but it does't really matter since its just the test

pyunifiprotect/api.py
```diff
@@ -723,6 +724,37 @@ async def get_camera_snapshot(
...
 
+    async def get_camera_video(
+        self, device_or_id: Union[Camera, str], start: datetime, end: datetime, channel: Union[CameraChannel, int] = 0
+    ) -> Optional[bytes]:
+        """Exports MP4 video from a given camera at a specific time"""
+
+        if isinstance(device_or_id, Camera):
+            camera = device_or_id
+        else:
+            camera = self.bootstrap.cameras[device_or_id]
```
bdraco, line 735: Can we avoid taking two different types here since this is new code?

bdraco (approved):
Installed and everything still works.  Did not test the new code path, only tested for collateral effects.

pyunifiprotect/api.py
AngellusMortis: The idea was to be as flexible as possible. You think it should only take the expected device itself?

bdraco: Ideally new functions don't have any `Union`ed arguments to minimize complexity.  Arguably this function isn't very complex, but they tend to grow so I try to always start new code with a very strict set of input requirements.

tests/conftest.py
AngellusMortis: I can look into it for a follow up PR. I am not still figuring async because it is complicated beast. If you can point too the primitives you had in mind. I would love to figure it out.

pyunifiprotect/api.py
AngellusMortis: That is fair enough. I will update the method calls then.

tests/conftest.py
bdraco: I think you could await `client.ws_task` with suppressing asyncio.CanceledError

bdraco (approved):
LGTM.  Any more cleanups or good to merge ?

AngellusMortis:
Nope, should be good for this PR.

PR #88, opened 2021-11-06
Simplifies creating test API clients

AngellusMortis:
Refactors @bdraco mentioned.

I am not sure why it is complaining about assigning to "Cannot assign to a method" now for the replacing the methods. I also cannot find a proper error code for that either to disable it via config since it is pretty common in tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import asyncio
 import base64
+import contextlib
 from datetime import datetime
 import json
 import os
@@ -7,7 +8,7 @@ from pathlib import Path
 from shlex import split
 from subprocess import run
 from tempfile import NamedTemporaryFile
-from typing import Any, Dict
+from typing import Any, Dict, Union
 from unittest.mock import AsyncMock, Mock
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


bdraco:
Where are you seeing the Cannot assign to a method?

tests/conftest.py
```diff
@@ -139,54 +142,43 @@ def ensure_debug():
...
-@pytest.mark.asyncio
-async def old_protect_client():
-    client = UpvServer(None, "127.0.0.1", 0, "username", "password")
-    client.api_request = AsyncMock(side_effect=mock_api_request)
-    client.api_request_raw = AsyncMock(side_effect=mock_api_request_raw)
-    client.ensure_authenticated = AsyncMock()
+async def setup_client(client: Union[ProtectApiClient, UpvServer], websocket: SimpleMockWebsocket):
+    client.api_request = AsyncMock(side_effect=mock_api_request)  # type: ignore
+    client.api_request_raw = AsyncMock(side_effect=mock_api_request_raw)  # type: ignore
+    client.ensure_authenticated = AsyncMock()  # type: ignore
```
AngellusMortis, line 148: Where I added the `type: ignore` comments in `setup_client`. I get them when I run mypy locally.

bdraco:
Thanks.  Will take a look.  May be tomorrow since I’m pretty wiped from travel.

tests/conftest.py
bdraco: Looks like its https://github.com/python/mypy/issues/6713

bdraco (approved):

PR #89, opened 2021-11-07
Adds update_device to API client + helper methods on devices

AngellusMortis:
With this PR, `ProtectApiClient` should have feature parity with `UvpServer`.


bdraco (approved):
LGTM 👍

PR #90, opened 2021-11-07
Adds set_speaker_volume to cameras

AngellusMortis:
@briis I split this out into a sperate PR to help you out adding new features to `pyunifiprotect`. We will need to able to control the output volume for the camera speakers for the `media_player` entities.

* Helper methods to update the device goes on the device class (in this case a camera).
    * I did a check for the feature flag that is required for the change to work (cannot change speaker volume if there are no speakers)
    * After you update the device class, call `.save_device` which will diff the object and then call `ProtectApiClient.update_device` which will do the `PATCH` to for the device like `UvpServer` did.
* You can write a test that verifies the `PATCH` request is what you expected from reverse engineering the change.
* Input validation is automatically done by `pydantic` via type hints 
    * For `speaker_settings.volume` it is a `PercentInt`, which is a `ContraintedInt`: https://pydantic-docs.helpmanual.io/usage/types/#strict-types for examples
    * See the test for what happens if you pass in invalid data.
    * A lot of the fields might still have Python base types (`int`, `str`, etc.). If you figure out validation for any additional fields, feel to make new types.

I am always free to help you figuring out if you need it!

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -279,6 +279,16 @@ async def camera_with_mic_obj(protect_client: ProtectApiClient):  # pylint: disa
 
 @pytest.fixture
 @pytest.mark.asyncio
+async def camera_with_speaker_obj(protect_client: ProtectApiClient):  # pylint: disable=redefined-outer-name
+    for c in protect_client.bootstrap.cameras.values():
+        if c.feature_flags.has_speaker:
+            return c
+
+    return None
+
+
+@pytest.fixture
+@pytest.mark.asyncio
 async def camera_with_privacy_obj(protect_client: ProtectApiClient):  # pylint: disable=redefined-outer-name
     for c in protect_client.bootstrap.cameras.values():
         if c.feature_flags.has_privacy_mask:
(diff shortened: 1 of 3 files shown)


tests/test_data_updates.py
```diff
@@ -314,6 +314,33 @@ async def test_camera_set_mic_volume(camera_with_mic_obj: Optional[Camera], leve
         )
 
 
+@pytest.mark.parametrize("level", [-1, 0, 100, 200])
```
AngellusMortis, line 317: `parametrize` will basically run this test this many times with different inputs. So you can test the boundary cases for the method.

tests/conftest.py
```diff
@@ -318,6 +318,16 @@ async def camera_with_mic_obj(protect_client: ProtectApiClient):  # pylint: disa
     return None
 
 
+@pytest.fixture
+@pytest.mark.asyncio
+async def camera_with_speaker_obj(protect_client: ProtectApiClient):  # pylint: disable=redefined-outer-name
```
AngellusMortis, line 323: pytest works using "fixtures" to setup data you need for a test. For all of these device update tests, I made a fixture to pull the exact device I wanted to test against to simplify the tests themselves.

Fixtures are kind of like classes and can "inherit" each other. `camera_with_speaker_obj` uses the `protect_client` fixture in this example.

https://docs.pytest.org/en/latest/how-to/fixtures.html

bdraco (approved):
LGTM

PR #91, opened 2021-11-07
Adds ability to play audio from URL to a camera via talkback settings

AngellusMortis:
* Adds ability to exclude fields from "changed" data for UFP objects (override `_get_excluded_changed_fields`)
* Adds `FfmpegCommand`: ffmpeg wrapper class to run ffmpeg commands. 
* Adds `TalkbackStream`: extends `FfmpegCommand` to play an input to a camera
* Adds `play_audio` to camera obj that wraps `TalkbackStream`

`play_audio`, `FfmpegCommand`, and `TalkbackStream` all obviously requires `ffmpeg` to use.

The way to test locally is to use the `unifi-protect shell` command:

```python
c = protect.bootstrap.cameras[id_of_camera_with_speaker]
await c.play_audio(path_to_mp3_file)
```

### DIFF
--- a/pyunifiprotect/stream.py
+++ b/pyunifiprotect/stream.py
@@ -1,0 +1,136 @@
+from __future__ import annotations
+
+from asyncio.subprocess import PIPE, Process, create_subprocess_exec
+from pathlib import Path
+from shlex import split
+from shutil import which
+from typing import TYPE_CHECKING, List, Optional
+
+from pyunifiprotect.exceptions import BadRequest, StreamError
+
+if TYPE_CHECKING:
+    from pyunifiprotect.data import Camera
+
+
+class FfmpegCommand:
+    ffmpeg_path: Path
+    args: List[str]
+    process: Optional[Process] = None
+
+    _stdout: Optional[List[str]] = None
+    _stderr: Optional[List[str]] = None
+
+    def __init__(self, cmd: str, ffmpeg_path: Optional[Path] = None) -> None:
+        self.args = split(cmd)
+
+        if "ffmpeg" in self.args[0]:
+            default_ffmpeg: Optional[str] = self.args.pop(0)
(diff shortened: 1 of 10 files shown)


pyunifiprotect/stream.py
```diff
@@ -0,0 +1,136 @@
...
+    _stdout: Optional[List[str]] = None
+    _stderr: Optional[List[str]] = None
+
+    def __init__(self, cmd: str, ffmpeg_path: Optional[Path] = None) -> None:
+        self.args = split(cmd)
+
+        if "ffmpeg" in self.args[0]:
+            default_ffmpeg: Optional[str] = self.args.pop(0)
+        else:
+            default_ffmpeg = which("ffmpeg")
```
bdraco, line 29: Technically this does I/O so it should probably run in the executor. If the disk is loaded the stats could block for a few seconds

bdraco (changes requested):

pyunifiprotect/stream.py
AngellusMortis: I added aioshutil and moved into `start`.

bdraco (approved):
LGTM. I did not test but since this is new functionality, I figure there isn't much risk of breaking anything

AngellusMortis:
I was able to confirmed it "works" with TTS. I went to HA and used the TTS field on one of my SONOS speakers:

image
https://home.wl.mort.is/api/tts_proxy/640ab2bae07bedc4c163f679a746f7ab7fb5d1fa_en-us_-_microsoft.mp3
Then I took the URL for the `media_content_id` attribute on the `media_player` (should look something like `{internal_url}/api/tts_proxy/{some_hex_id}_{lang}_-_{tts_provider}.mp3`). I was able to fee that TTS URL directly into to `play_audio`

PR #93, opened 2021-11-07
Updates readme

AngellusMortis:


AngellusMortis:
@briis how does it look? Added some status badges, a install/usage section for the main API and some copy changes.

briis:
Looks very good. Gives a good overview and basic usages.  Once  `unifiprotect` has been migrated we can point to that for more examples on usage.

PR #99, opened 2021-11-15
Workflow to test against live data

AngellusMortis:
I cleaned up the workflow I have been running on my personal fork, made some docs and decided to made a PR to merge it into the main repo. 

PR is in draft for now, but it is 100% done. There is some stuff that needs to be added to the repo itself that I am coordinating with @briis to see if we can add it.


AngellusMortis:
Okay, we got the secrets and the new runner set up so my workflow should work after the PR is merged. @bdraco did you want to review? Or are you okay with me merging?

bdraco (approved):

Issue #98, opened 2021-11-13
Adds Smart Track for Smart Detection events


AngellusMortis:
I figured out you can get the smart detect data for an event at `/proxy/protect/api/events/{event_id}/smartDetectTrack`. I would like to use this to build in new features for smart detection events.

* New class for Smart Detect Tracks
* New API client endpoint / `get_` on `Event` to get the smart detect tracks (cached on event after first call)
* Helper method on Event to get the triggering zones for smart detection
* Helper method on Event to take first frame of smart detect track and render rectangle for smart detect track (as an alternative to the event thumbnail) Result will look something like this: 
![image](https://user-images.githubusercontent.com/490848/141654096-f7699265-c64e-4c24-858e-fd70f8c3219d.png)

PR #101, opened 2021-11-15
Adds retrieving the Smart Detect Track for a smart detection

AngellusMortis:
Implements https://github.com/briis/pyunifiprotect/issues/98. Everything except the last point about creating the image. 
The `ts=` query param for `camera/{id}/snapshot` did not work like I expected. See https://github.com/briis/pyunifiprotect/pull/100. The only other way I can think to implement that is to splice frames out of the video using ffmpeg, but the exported videos are not 100% accurate, so it could result into bad images.

*  `ProtectApiClient`: Adds passing API reference for `get_device` methods
*  `ProtectApiClient`: Adds `get_event` method
*  `ProtectApiClient`: Adds `get_event_smart_detect_track_raw` / `get_event_smart_detect_track` methods
* `.data`: Adds `SmartDetectTrack` and `SmartDetectItem` classes
* `Event`: Adds `get_smart_detect_track` and `get_smart_detect_zones`

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -27,6 +27,7 @@ from pyunifiprotect.data import (
     ModelType,
     ProtectModel,
     Sensor,
+    SmartDetectTrack,
     Viewer,
     WSPacket,
     WSSubscriptionMessage,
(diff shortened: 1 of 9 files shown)


bdraco:
Will try to take a look tomorrow.  I've been in flight for the better part of the day.

bdraco (approved):
Looks good.  Did some basic testing for side effects. 👍

PR #104, opened 2021-11-25
Fixes cookies and SSL validation

AngellusMortis:
Couple of bugs I came across from testing

* `is_authenticated` was not working with the default session because `TOKEN` was a `httponly` cookie
* `verify_ssl=True` was also not working and resulting in SSL cert to never be verified because if any bool = no verify
* Adds semantic version parsing for NVR version to make it easier to compare versions

## HA Install Link:

```
git+https://github.com/briis/pyunifiprotect.git@http-fixes#pyunifiprotect==1.0.5
```

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -12,7 +12,7 @@ from urllib.parse import urljoin
 from uuid import UUID
 
 import aiohttp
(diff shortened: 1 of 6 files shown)


bdraco (approved):
Tested with `unifiprotect` master. No unexpected side effects

PR #105, opened 2021-11-26
Requires no HDR to set WDR

AngellusMortis:
I am kind of torn on this one. I _think_ this is a new change in Protect that does not expose changing WDR when the camera has HDR. I have 1x G4 Bullet (HDR), 4x G4 Pros (HDR), and 1x G4 Doorbell (no HDR) and WDR _only_ appears on my G4 Doorbell now, in the Protect WebUI.

I remember changing the WDR for my G4 Pros previously, but it seems to have changed. It looks like changing WDR on an HDR camera does still have some effect (I see the lens do something, but the picture quality does not really change). So we may not want to enforce this, but if we want to stay inline with how Protect works...

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -761,6 +761,10 @@ class Camera(ProtectMotionDeviceModel):
 
     async def set_wdr_level(self, level: WDRLevel) -> None:
         """Sets WDR (Wide Dynamic Range) on camera"""
+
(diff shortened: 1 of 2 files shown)


bdraco (approved):
LGTM

AngellusMortis:
I did figure this out (why I thought it was new). Setting the Video Mode to `highfps` disables HDR on cameras.

PR #107, opened 2021-11-26
Adds option to force emit messages when an object is changed

AngellusMortis:
* Adds option to force emit messages when an object is changed
* Ensures PATCH requests for updating an LCDMessage object is always what UFP is expecting:
   * If no `resetAt` was passed in, it was defaulted to +1 minute
   * If no `type` was passed in (for CUSTOM_MESSAGE -> CUSTOM_MESSAGE changes), the display would break

This one is to "fix" a bug in UFP. When you update a Viewer, it does not emit a WS message like updating other objects do. This will let you to pass `force_emit=True` on `save_device` to construct a fake WS message of the change.

This will allow downstream subscribers (like Home Assistant) to trigger updates properly in the absence of a real UFP message.

Other bundled changes/fixes for HA integration:
* `ProtectApiClient`: Ensure `start` param passed into `get_events` inside of `update` is always set
* `Bootstrap`: Simplifies event processing and only captures the reference to the latest event IDs instead of trying to update the object. Other WS messages will do that.
* Adds `set_sensitivity` and `set_duration` helper methods for lights

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -18,7 +18,12 @@ from typing import (
 from pydantic import BaseModel
 from pydantic.fields import SHAPE_DICT, SHAPE_LIST, PrivateAttr
 
-from pyunifiprotect.data.types import ModelType, StateType
+from pyunifiprotect.data.types import ModelType, ProtectWSPayloadFormat, StateType
+from pyunifiprotect.data.websocket import (
+    WSJSONPacketFrame,
+    WSPacket,
+    WSPacketFrameHeader,
+)
 from pyunifiprotect.exceptions import BadRequest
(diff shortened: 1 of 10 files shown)


pyunifiprotect/data/base.py
```diff
@@ -587,6 +596,23 @@ async def save_device(self) -> None:
         await self.api.update_device(self.model, self.id, updated)
         self._initial_data = new_data
 
+        if force_emit:
```
bdraco, line 599: You could save some indent by reversing the condition and returning

AngellusMortis: Good call.

bdraco (approved):

PR #108, opened 2021-11-26
Adds various new helper methods

AngellusMortis:
* `ProtectApiClient`: Adds `update_nvr` and `reboot` methods to update main NVR object and reboot adoptable devices
* `ProtectDeviceModel`: Adds `set_ssh` and `reboot` helper methods
* Moves `save_device` from `ProtectAdoptableDevice` to `ProtectModelWithId` so it can be used for NVR objects
* `Camera`: Adds `DEFAULT` option for `set_lcd_text` to be able to set the `reset_at` value to the `default_message_reset_timeout`
* `NVR`: Adds `set_default_reset_timeout`, `set_default_doorbell_message`, `add_custom_doorbell_message`, `remove_custom_doorbell_message` methods to manage Doorbell message settings.
* `FfmpegCommand`: Fixes capturing stdout/stderr and adds more logging


briis:
What is the purpose of enabling ssh for the Camera? What do people use this for?

AngellusMortis:
Tha main use case is it allows you to edit files on the camera which is needed to be able to change your chime audio for the G4 Doorbell.

briis:
Interesting. But that would need us to create a service or similar in Home Assistant for this. to be used? And this is not even possible from the UI in Protect as I can see.

AngellusMortis:
I am not suggesting we build this specific thing out further. _Maybe_ a switch in HA that is disabled by default, but beyond that, it is more of an "advanced" feature for someone wanting to use it. Largely outside of HA.

AngellusMortis:
Making this one draft as I have a few new features I discovered I am going to add into this one.

bdraco (approved):

PR #118, opened 2021-12-02
Optmizations for `ProtectApiClient`

AngellusMortis:
* Increases Bootstrap poll interval to 15 minutes if WS is activate. Does not need to be polled every 60 seconds
* Adds arguments to ignore a number of events from WS
* Adds test to benchmark processing Websocket packets
* Optimizes `update_from_dict` to not call `.copy` or `.construct`.

No optimization:
image

With optimizations: 
image

second pass optimizations: 
image

The `test_ws_filtered` is closer to what the HA performance gains should be. The only difference there is that NVR messages were ignored for that test to improve coverage. 

About a ~35% performance increase on the mean process time for the `update_from_dict` optimizations and then an additional 25% (52% total) increase filtering out WS messages.

### DIFF
--- a/pyunifiprotect/data/bootstrap.py
+++ b/pyunifiprotect/data/bootstrap.py
@@ -22,6 +22,7 @@ _LOGGER = logging.getLogger(__name__)
 
 MAX_SUPPORTED_CAMERAS = 256
 MAX_EVENT_HISTORY_IN_STATE_MACHINE = MAX_SUPPORTED_CAMERAS * 2
+STATS_KEYS = {"storageStats", "stats", "systemInfo"}
 
 EVENT_ATTR_MAP: Dict[EventType, Tuple[str, str]] = {
     EventType.MOTION: ("last_motion", "last_motion_event_id"),
@@ -95,7 +96,12 @@ class Bootstrap(ProtectBaseObject):
 
         self.events[event.id] = event
 
-    def process_ws_packet(self, packet: WSPacket) -> Optional[WSSubscriptionMessage]:
+    def process_ws_packet(
+        self, packet: WSPacket, models: Optional[Set[ModelType]] = None, ignore_stats: bool = False
+    ) -> Optional[WSSubscriptionMessage]:
+        if models is None:
+            models = set()
+
         if not isinstance(packet.action_frame, WSJSONPacketFrame):
             _LOGGER.debug("Unexpected action frame format: %s", packet.action_frame.payload_format)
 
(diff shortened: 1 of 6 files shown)


pyunifiprotect/data/bootstrap.py
```diff
@@ -147,6 +164,15 @@ def process_ws_packet(self, packet: WSPacket) -> Optional[WSSubscriptionMessage]
...
                 )
             if model_type in ModelType.bootstrap_models() or model_type == ModelType.EVENT.value:
+
+                if model_type == ModelType.CAMERA.value and ignore_stats:
+                    for key in list(data.keys()):
+                        if key in STATS_KEYS:
+                            del data[key]
+                    # nothing left to process
+                    if data == {}:
+                        return None
```
bdraco, line 174: Similar code to line 148, could be a function

AngellusMortis: Was thinking about it, but not sure since it was only 3 lines (cannot really break out the return if).

bdraco: Maybe 
```

if _prune_stats_keys(data):
    return None

....

def _prune_stats_keys(data):
    if STATS_KEYS.intersection(data):
        for key in list(data):   
            if key in STATS_KEYS:
                del data[key]
    # nothing left to process
    return data == {}
```

AngellusMortis: I refactored it a bit. There are only 3 keys in that set, so I do not think getting fancy with set logic will give us any real boost here. The number of keys from the WS messages is also pretty limited as well (~3 max top level, ~6 max if you include the nested stats keys that are already being stripped out as a result).

AngellusMortis: Added it anyways. haha.

bdraco:
Screen Shot 2021-12-02 at 12 42 27


Looks like this now

AngellusMortis:
So that is a pretty good improvement. That is without filtering any of the messages as well too though since that requires a change to how we initialize the client on the HA side.

AngellusMortis:
Added a commit on the integration branch to use the new options to API Client.

bdraco:
Screen Shot 2021-12-02 at 15 43 00

bdraco:
On another instance

Screen Shot 2021-12-02 at 15 49 24

bdraco:
Testing on the original instance  https://github.com/briis/unifiprotect/pull/354#issuecomment-984219175

Original was ~47% of the event loop cpu time
Optimized version is ~25% of the event loop cpu time

bdraco (approved):

bdraco:
Hopefully the filtering will get us back closer to the original implementation performance which was between 4-6%  on that instance

PR #119, opened 2021-12-02
Reduce iterations

bdraco:


AngellusMortis:
I was just going through and doing this as well. I was going to redo all of the for loops (there are a lot) to use the set logic.

PR #122, opened 2021-12-03
Refactors packet processing

AngellusMortis:


pyunifiprotect/data/bootstrap.py
```diff
@@ -151,6 +151,94 @@ def _create_stat(self, packet: WSPacket, keys_set: List[str], filtered: bool) ->
...
+    ) -> Optional[WSSubscriptionMessage]:
+        data = _remove_stats_keys(data, ignore_stats)
+        # nothing left to process
+        if len(data) == 0:
+            self._create_stat(packet, [], True)
+            return None
+
+        data = self.nvr.unifi_dict_to_dict(data)
+        old_nvr = self.nvr.copy()
+        self.nvr = self.nvr.update_from_dict(deepcopy(data))
```
bdraco, line 193: Any way to get rid of this `deepcopy`?  
Probably would require more refactoring in `update_from_dict` though since it is mutated there.

AngellusMortis: Yeah, I do not think I could rid of those two. The data dictionary is mutated a lot using update/creation.

But the other two deepcopy instances are both _after_ the packet has already been filtered. So, the impact is much less.

bdraco:
Screen Shot 2021-12-03 at 08 53 05

`deepcopy` effectively gone from the profile

bdraco (approved):
With this change, down to ~2-4% of the event loop time on the original instance.  So this is as fast or better as the original implementation 👍

PR #123, opened 2021-12-04
Creates release workflow

AngellusMortis:
This workflow needs a PyPi API token secret added to use.

1. Go to https://pypi.org/manage/account/token/
2. Create a token for "Project: pyunifiprotect"
3. Go to https://github.com/briis/pyunifiprotect/settings/secrets/actions
4. Add "New repository secret" named `PYPI_API_TOKEN` with the value from above

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,41 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [ published ]
+
+jobs:
+  deploy:
+    name: Publish to PyPi
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check repo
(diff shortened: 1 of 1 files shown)


briis:
This is now done. Hope it works.

AngellusMortis:
I copied and pasted it from one that I knew was working. Only difference is `build` instead of `flit` for building the package. Same upload process though. I will merge it and test it out once @bdraco gets a chance to look at https://github.com/briis/pyunifiprotect/pull/124 and merge it.

briis:
If I wanted to use this for some of the other projects I have, would I have to change anything, except for the token reference?

AngellusMortis:
You _might_ need to change steps 3 and 4. Since [PEP 517](https://www.python.org/dev/peps/pep-0517/), there is a standard for making builders. So, you can use whatever builder you want now. Step 3 is "install requirements for builder" (in this case `build`, but I also mentioned I used `flit` on another package of mine). Step 4 is "build the package".

AngellusMortis:
Here is the other project I pulled it from. As you can see, it is identical except for steps 3/4 https://github.com/AngellusMortis/sxm-player/blob/master/.github/workflows/release.yml

Then here is the Github template for uploading a package on release: https://github.com/briis/pyunifiprotect/new/master?filename=.github%2Fworkflows%2Fpython-publish.yml&workflow_template=python-publish

That is what I originally based mine off of.

AngellusMortis:
It worked. https://pypi.org/project/pyunifiprotect/1.2.1/

PR #127, opened 2021-12-08
Adds retry logic for thumbnails/heatmaps

AngellusMortis:


pyunifiprotect/api.py
```diff
@@ -884,6 +886,20 @@ async def get_camera_video(
...
+    async def _get_image_with_retry(self, path: str, **kwargs: Any) -> Optional[bytes]:
+        """Retries image request until it returns or timesout. Used for event images like thumbnails and heatmaps"""
+
+        now = time.monotonic()
+        timeout = now + RETRY_TIMEOUT
+        data: Optional[bytes] = None
+        while data is None and now < timeout:
+            data = await self.api_request_raw(path, raise_exception=False, **kwargs)
+            if data is None:
+                await asyncio.sleep(0.5)
```
bdraco, line 898: Should we increase the sleep window each loop and do `min(window, reminaing)`?

AngellusMortis: That would reduce the number of calls to the NVR, but make it take longer for the thumbnail to come back. If it takes 5 seconds for it to generate, but we are doing the longer windows, it could take 8 seconds to come back (if that makes sense).

bdraco (approved):

PR #130, opened 2021-12-11
Fixes raise exception

AngellusMortis:
The `raise_exception` argument passed into the `api_request*` methods was not actually working because higher up the chain it was checking the status code first. 

This was causing the retry logic for the event thumbnails to fail.


AngellusMortis:
Tested in my prod instance with the blueprints I have been working on. This works great now.

bdraco (approved):

PR #131, opened 2021-12-12
Adds option to override connection host

AngellusMortis:
Let's users match their `connection_host` to their `host`. Would allow access RTSP streams over the Internet.


bdraco (approved):

bdraco:
Looks good.  Reviewed on my mobile so couldn't test but simple enough

PR #132, opened 2021-12-18
Fixes bitrate for talkback

AngellusMortis:
This gets talkback to work on G4 Doorbell and G4 Dome. Looks like the issue was setting the quality.


AngellusMortis:
I think this is correct. If you happen to understand more about sampling/bit rates then me, let me know if my math is wrong.

AngellusMortis:
For sure have the correct bit rate now. I decompiled the Java app and 48k is what it uses.

bdraco (approved):

PR #133, opened 2021-12-20
Updates for Protect 1.21-beta1

AngellusMortis:
* `ProtectAdoptableDeviceModel`: Adds `market_name` attribute
* `EventType`: Adds `lightMotion`, `sensorMotion`, and `recordingModeChanged` events
* `Bootstrap`: Adds processing for new Flood Light and Smart Sensor events
* `Camera`: Adds `is_wireless_uplink_enabled`
* `Sensor`: Adds `last_motion_event_id` / `last_motion_event`
* `Event` Adds parsing of `metadata` to get sensor/light objects from motion events
* `Event`: Adds `timestamp`
* `ProtectApiClient`: Use newer thumbnail/heatmap URL (`events/{event_id}/{thumbnail|heatmap}`). I _think_ this one generates in real time and does _not_ require the event to actually end to get an image back for (will be tested with HA after release).
* Generates new 1.21-beta1 sample data

Filter to exclude the sample data files from view: https://github.com/briis/pyunifiprotect/pull/133/files?file-filters%5B%5D=.py


AngellusMortis:
@kasper79

If you could install this branch (`pip install -U git+https://github.com/briis/pyunifiprotect.git@1.21`) and then [generate sample data / runs the test suite](https://github.com/briis/pyunifiprotect#generating-and-testing-with-real-data), that would be great. Make sure you trigger any UP Sense events you can find while the WS capture is happening.

pyunifiprotect/data/nvr.py
```diff
@@ -82,6 +82,63 @@ def event(self) -> Optional[Event]:
...
+    @classmethod
+    def _get_unifi_remaps(cls) -> Dict[str, str]:
+        return {
+            **super()._get_unifi_remaps(),
+            "from": "fromValue",
+            "to": "toValue",
+        }
+
+    @classmethod
+    def unifi_dict_to_dict(cls, data: Dict[str, Any]) -> Dict[str, Any]:
```
bdraco, line 106: Can you use an intersection + loop here?

pyunifiprotect/data/nvr.py
```diff
@@ -82,6 +82,63 @@ def event(self) -> Optional[Event]:
...
+
+    def unifi_dict(self, data: Optional[Dict[str, Any]] = None, exclude: Optional[Set[str]] = None) -> Dict[str, Any]:
+        data = super().unifi_dict(data=data, exclude=exclude)
+
+        # all metadata keys optionally appear
+        for key, value in list(data.items()):
+            if value is None:
+                del data[key]
+
+        if "lightId" in data:
```
bdraco, line 128: Can you use an intersection + loop here?

bdraco (commented):
This looks great.  I'll be home in a bit and can test.

bdraco:
Did some basic testing.  Everything seems ok

bdraco (approved):

PR #134, opened 2021-12-24
Fixes pre-1.20 record mode error

AngellusMortis:
Not really easy to test this one, but hopefully the new unit test should handle it.

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -46,6 +46,7 @@ from pyunifiprotect.data import (
     WSSubscriptionMessage,
     create_from_unifi_dict,
 )
+from pyunifiprotect.data.types import RecordingMode
 from pyunifiprotect.exceptions import BadRequest, NotAuthorized, NvrError
 from pyunifiprotect.utils import (
     get_response_reason,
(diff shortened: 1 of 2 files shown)


pyunifiprotect/api.py
```diff
@@ -701,6 +702,18 @@ async def get_bootstrap(self) -> Bootstrap:
         """
 
         data = await self.api_request_obj("bootstrap")
+        # fix for UniFi Protect bug, some cameras may come back with and old recording mode
+        # "motion" and "smartDetect" recording modes was combined into "detections" in Protect 1.20.0
```
bdraco, line 706: Is this going to be wrong again the next websocket update ?  Should we change this where the data is parsed?

AngellusMortis: No, if you check out the [reported issue](https://github.com/briis/unifiprotect/issues/427), I had the user goes through and switch all of the recording modes for their cameras manually and it "solved" the issue. This is essentially doing the same thing. 

I do not know why, but it looks like some years cameras where not migrated to the new recording mode type. This just automates the process of having them do so we do not need to keep the old ones (which will bubble up and cause headache in HA since they are not valid selectable options for the Select Entities).

bdraco (approved):

PR #137, opened 2021-12-28
Play audio without blocking until completion

AngellusMortis:
I realized that HA never updates the speaker state because UniFi Protect does not send WS messages about talkback. This will let us start playing audio, update speaker state and then block until completion so we can update speaker state again.

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -1021,7 +1021,7 @@ class Camera(ProtectMotionDeviceModel):
         self.talkback_stream = TalkbackStream(self, content_url, ffmpeg_path)
         return self.talkback_stream
 
-    async def play_audio(self, content_url: str, ffmpeg_path: Optional[Path] = None) -> None:
+    async def play_audio(self, content_url: str, ffmpeg_path: Optional[Path] = None, blocking: bool = True) -> None:
         """
         Plays audio to a camera through its speaker.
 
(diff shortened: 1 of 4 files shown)


pyunifiprotect/data/devices.py
```diff
@@ -1031,9 +1031,21 @@ async def play_audio(self, content_url: str, ffmpeg_path: Optional[Path] = None)
...
+        * `blocking`: Awaits stream completion and logs stdout/stderr
         """
 
         stream = self.create_talkback_stream(content_url, ffmpeg_path)
+        await stream.start()
+
+        if blocking:
+            await self.wait_until_audio_completes()
+
+    async def wait_until_audio_completes(self) -> None:
```
bdraco, line 1043: Should this have a timeout?

AngellusMortis: In this case, probably not. It would be hard to create a timeout since the file could have a variable length. If it is a 5 second audio file, the timeout would probably be ~6 seconds. If it was 30, probably ~31 seconds.

The intention is just so you can know when it is complete and come back and do something later (in the case of HA, trigger another state update).

AngellusMortis: There is probably some weird cases where it _could_ hang forever, like zombie processes, but those should be pretty rare, if ever (I do not think I have ever seen ffmpeg have a zombie process). As long as ffmpeg exists, either successfully or not, this will eventually do something.

bdraco (approved):

PR #141, opened 2022-01-04
Adds more features for Smart Sensor

AngellusMortis:
TODO: More tests

* `Event`/`Sensor`: Adds support for `sensorOpened`, `sensorClosed`, `sensorExtermeValue`, and `sensorAlarm` events
* `Sensor`: Adds better enum typing
* `Sensor`: Adds various helper methods to configure sensor
* `ProtectApiClient`: For `get_events_raw`, replace `camera_ids` with `types` to filter based on event type.
* Generates new sample data with Sense devices

Only thing really missing for the UP Sense after this leak events (leak electrodes have not been released yet)

Filter to exclude sample files: https://github.com/briis/pyunifiprotect/pull/141/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.ini&file-filters%5B%5D=.py&file-filters%5B%5D=.txt


bdraco (approved):
LGTM.  I do not have a smart sensor so I cannot validate with production testing

PR #142, opened 2022-01-04
Removes deprecated unifi_ code. Drop 3.8 support. Upgrade mypy to pyproject.toml

AngellusMortis:
Now that pydantic _finally_ released 1.9.0, now is a good time to do a bunch of upgrades. Starting with 

* remove `UpvServer` and any remaining code that uses it
* remove Python 3.8 support
* replace `mypy.ini` with config in `pyproject.toml`


bdraco:
I might be a bit early to drop 3.8 since core dev still supports it

https://github.com/home-assistant/core/blob/dev/homeassistant/const.py#L13

AngellusMortis:
It is suppose to be removed soon. It was originally going to be 2022.1,but that release was canceled.

https://www.home-assistant.io/blog/2021/11/03/release-202111/

AngellusMortis:
Oh, found the drop PR to remove support: https://github.com/home-assistant/core/pull/62340

bdraco:
If we need to do a release of this before 2022.2 it would put us in a bit of a bind since there may still be users using 3.8

bdraco (approved):

PR #144, opened 2022-01-06
Fixes anonymizing data. Add logger support

AngellusMortis:


pyunifiprotect/cli.py
```diff
@@ -113,7 +113,7 @@ def generate_sample_data(
         output_folder = (tests_folder / "sample_data").absolute()
 
     protect = ProtectApiClient(address, port, username, password, verify_ssl=verify, debug=True)
-    SampleDataGenerator(protect, output_folder, anonymize, wait_type).generate()
+    SampleDataGenerator(protect, output_folder, anonymize, wait_type, cli=True).generate()
```
bdraco, line 116: Can we avoid making this aware of the cli?

Can we give this a progress callable instead?

AngellusMortis: This might only be a VS Code issue, but `typer` is actually based on `click`, which is already supposed to do that. The progress bar is not even supposed to print if `click` detects it is not a terminal.

AngellusMortis: But I definitely can refactor it to remove refernces of typer in the `SampleDataGenerator` and just pass in callables.

bdraco (approved):

PR #149, opened 2022-01-07
Improves is_ attrs

AngellusMortis:
After testing more in my main HA install, I found some more issues. 

Issues:

* `emit_message` was not being awaited
* As a result of above ^ the triggers were very inconsistent

Changes:

* Use fixed point in future as when the event "ends". This let's look at the WS value for the 30 second cutoff, but still have the 3 second "auto-expire"
* No more datetime math for the `is_` property. 
* Improves WS reconnects in the event of a disconnect

Since I am kind of tired of making releases only to test in my prod install. I am going to directly install this branch in my prod HA and test it further before marking as read.

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -1,6 +1,7 @@
 """Unifi Protect Data."""
 from __future__ import annotations
 
+import asyncio
 from datetime import datetime, timedelta
 from ipaddress import IPv4Address
 from typing import (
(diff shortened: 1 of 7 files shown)


pyunifiprotect/data/base.py
```diff
@@ -603,6 +606,10 @@ def unifi_dict_to_dict(cls, data: Dict[str, Any]) -> Dict[str, Any]:
 
         return super().unifi_dict_to_dict(data)
 
+    def _event_callback_ping(self) -> None:
```
bdraco, line 609: Do we need to cancel and reschedule this if another message of the same type is emitted within the interval?

AngellusMortis: I mean we could, but all it does is send an empty packet as a ping.

AngellusMortis: I was sending the _at timestamp, but stopped for that reason. The ping could come in after another event.

bdraco (approved):

PR #151, opened 2022-01-10
Makes IO async friendly. Adds profile_ws command. Add zip option to data gen

AngellusMortis:


pyunifiprotect/utils.py
```diff
@@ -311,3 +320,43 @@ def print_ws_stat_summary(stats: List[WSStat], output: Optional[Callable[[Any],
...
+
+    if output_path:
+        json_data = [s.__dict__ for s in protect.bootstrap.ws_stats]
+
+        def write() -> None:
+            if output_path:
+                with open(output_path, "w", encoding="utf8") as outfile:
+                    json.dump(json_data, outfile, indent=4)
+
+        _LOGGER.debug("Writing json to %s...", output_path)
```
bdraco, line 357: Looks like similar code to `write_json_file`

bdraco (approved):

pyunifiprotect/utils.py
AngellusMortis: It is, but completely different file. I could refactor this out a bit though.

PR #161, opened 2022-01-13
Refactor Websocket code

AngellusMortis:
This PR will be a breaking change and need a major version bump (3.0) as it changes a number of public access variables on `ProtectApiClient`. Even if they essentially still function the same. 

Completely refactors and redesigns Websocket code:

* All Websocket code is not pulled out into a new Websocket class
* Adds timeout to handle Websocket connection getting "stuck"
* Decreases reconnect time to 10 seconds
* Simplifies async logic: WS task and WS session are no longer tracked on the class level. The disappear as soon as they are no longer relevant.
* Simplifies `last_update_id` and ensures it is always updated for future reconnects

Unrelated: Fixes python-xdist running with using "run tests", but _not_ running when you "debug tests".

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -104,6 +104,7 @@ class BaseApiClient:
         if self._port != 443:
             url += f":{self._port}"
 
+        url += self.ws_path
         last_update_id = self._get_last_update_id()
         if last_update_id is None:
             return url
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -400,9 +400,9 @@ class ProtectApiClient(BaseApiClient):
     _minimum_score: int
     _subscribed_models: Set[ModelType]
     _ignore_stats: bool
+    _ws_subscriptions: List[Callable[[WSSubscriptionMessage], None]]
     _bootstrap: Optional[Bootstrap] = None
     _last_update_dt: Optional[datetime] = None
-    _ws_subscriptions: List[Callable[[WSSubscriptionMessage], None]] = []
     _connection_host: Optional[Union[IPv4Address, str]] = None
 
     ignore_unadopted: bool
@@ -434,6 +434,7 @@ class ProtectApiClient(BaseApiClient):
         self._minimum_score = minimum_score
         self._subscribed_models = subscribed_models or set()
         self._ignore_stats = ignore_stats
+        self._ws_subscriptions = []
         self.ignore_unadopted = ignore_unadopted
 
         if override_connection_host:
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -71,7 +71,7 @@ class BaseApiClient:
     _websocket: Optional[Websocket] = None
 
     api_path: str = "/proxy/protect/api/"
-    ws_path: str = "/proxy/protect/ws/"
+    ws_path: str = "/proxy/protect/ws/updates"
 
     def __init__(
         self,
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -58,12 +58,6 @@ RETRY_TIMEOUT = 10
 _LOGGER = logging.getLogger(__name__)
 
 
-def _reset_connection(connection: Optional[Union[aiohttp.ClientSession, aiohttp.ClientWebSocketResponse]]) -> None:
-    if connection is not None and not connection.closed:
-        loop = asyncio.get_running_loop()
-        loop.create_task(connection.close())
-
-
 class BaseApiClient:
     _host: str
     _port: int
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -44,6 +44,7 @@ from pyunifiprotect.utils import (
     to_js_time,
     utc_now,
 )
+from pyunifiprotect.websocket import Websocket
 
 NEVER_RAN = -1000
 # how many seconds before the bootstrap is refreshed from Protect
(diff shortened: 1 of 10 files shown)


pyunifiprotect/api.py
```diff
@@ -323,13 +323,17 @@ def reset_ws(self) -> None:
...
-                    self._ws_session = None
+                self._ws_task = None
             except Exception:  # pylint: disable=broad-except
                 _LOGGER.exception("Could not cancel WS task")
 
+        if self._ws_task is None:
+            self._ws_connection = None
+            if self._ws_session is not None:
+                _LOGGER.debug("Reseting WS session...")
+                self._ws_session.close()
```
bdraco, line 334: Looks like this was being awaited before?

AngellusMortis: Oh crap. It was. maybe that is the problem!

pyunifiprotect/api.py
```diff
@@ -57,6 +57,12 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+def _reset_connection(connection: Optional[Union[aiohttp.ClientSession, aiohttp.ClientWebSocketResponse]]) -> None:
+    if connection is not None and not connection.closed:
+        loop = asyncio.get_running_loop()
+        loop.create_task(connection.close())
```
bdraco, line 63: This looks like it's race prone.  Is there a chance it won't be closed before the next one is started?

bdraco:
Maybe it would be easier if there was a separate function to start the websocket and the watchdog task instead of using the periodically polling?

AngellusMortis:
> Maybe it would be easier if there was a separate function to start the websocket and the watchdog task instead of using the periodically polling?

I think that is a good future enhancement once we get the polling to work reliably. Adding another task would complicate the process and make it harder to debug and get working better.

pyunifiprotect/api.py
AngellusMortis: It is, but I do not believe that to be a problem. My understanding is that you are supposed to close the connections to prevent resource leaks. So as long as the connect is eventually closed, it should not matter if it happens before another attempt is made.

Unless I am missing something on how it async connections work.

bdraco:
Aaron had similar issues with aioambient

I think this was the final solution https://github.com/bachya/aioambient/blob/dev/aioambient/websocket.py

pyunifiprotect/api.py
bdraco: It's not a problem unless the protect device will start to get upset about how many connections there are to the websocket

AngellusMortis:
Hmmm... I still see what I can do to rework the Websocket stuff.

AngellusMortis:
Okay. I complete refactored out and redesigned the Webocket code so it is not entangled in the API client.

pyunifiprotect/websocket.py
```diff
@@ -0,0 +1,136 @@
...
+            self._timer_task.cancel()
+
+    async def _timeout(self) -> None:
+        _LOGGER.debug("WS timed out")
+        if not await self.reconnect():
+            await self._timeout()
+
+    async def _reset_timeout(self) -> None:
+        _LOGGER.debug("WS timeout reset")
+        self._cancel_timeout()
```
bdraco, line 92: The unifi protect websocket is a lot busier than most. Canceling and readding timers is going to be a bit expensive. 

We usually set a value to the time.monotonic and have something that checks it periodically to see if we hit the timeout/failure stare

AngellusMortis:
Testing this out now. It looks like messages are not bubbling up to HA. So, I am debugging that issue now. A bunch of minor changes likely inbound.

pyunifiprotect/websocket.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _cancel_timeout(self) -> None:
+        if self._timer_task:
+            self._timer_task.cancel()
+
+    async def connect(self) -> bool:
+        """Connect the websocket."""
+
+        _LOGGER.debug("Scheduling WS connect...")
+        asyncio.create_task(self._websocket_loop())
```
bdraco, line 116: Pass in a `asyncio.Future` or `asyncio.Event` and set it once connected so you can await it wrapped in `asyncio.wait_for` without a timeout below without the sleep.

AngellusMortis: Oh neat. I was wondering a good way to do that. Could not fix it out.

pyunifiprotect/websocket.py
```diff
@@ -0,0 +1,162 @@
...
+    _headers: Optional[Dict[str, str]] = None
+    _timer_task: Optional[asyncio.Task[None]] = None
+    _ws_connection: Optional[ClientWebSocketResponse] = None
+
+    def __init__(
+        self,
+        url: str,
+        auth_callback: CALLBACK_TYPE,
+        timeout: int = 30,
+        reconnect_wait: int = 10,
```
bdraco, line 31: Maybe call this a backoff?

pyunifiprotect/websocket.py
```diff
@@ -0,0 +1,162 @@
...
+    def _cancel_timeout(self) -> None:
+        if self._timer_task:
+            self._timer_task.cancel()
+
+    async def connect(self) -> bool:
+        """Connect the websocket."""
+
+        start_event = asyncio.Event()
+        _LOGGER.debug("Scheduling WS connect...")
+        asyncio.create_task(self._websocket_loop(start_event))
```
bdraco, line 119: Should we store the task and cancel the task as well in the event we timeout?

AngellusMortis: Closing the Websocket connect causes it to clean itself up with the `finally` clause on inside the loop. Canceling it could cause it to not shutdown the connection cleanly.

bdraco (approved):
Did some testing with this and everything appears to be working as expected

bdraco:
Also confirmed the integration could be reloaded successfully

PR #164, opened 2022-01-17
More Websocket improvements

AngellusMortis:
* Force re-auth on multiple reconnect failures
* Log Websocket exceptions
* Ensure only a single connection attempt is ever made
* Improves disconnection/reconnect log messages
* Fix timeout if Websocket connect fails

### DIFF
--- a/pyunifiprotect/websocket.py
+++ b/pyunifiprotect/websocket.py
@@ -8,6 +8,7 @@ from aiohttp import ClientSession, ClientWebSocketResponse, WSMessage, WSMsgType
 
 _LOGGER = logging.getLogger(__name__)
 CALLBACK_TYPE = Callable[..., Coroutine[Any, Any, Optional[Dict[str, str]]]]
+RECENT_FAILURE_CUT_OFF = 30
 
 
 class Websocket:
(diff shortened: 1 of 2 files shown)


AngellusMortis:
Been running this one a little over a day in my prod instance already. Looks like it fixes various of issues from the previous PR. I have not seen re-connect failures since (still maybe take more than a day of running to verify and I have been doing some network maintenance today as well causing restarts for HA).

pyunifiprotect/websocket.py
```diff
@@ -64,25 +69,51 @@ def _process_message(self, msg: WSMessage) -> bool:
...
         try:
+            self._ws_connection = await session.ws_connect(self.url, ssl=self.verify, headers=self._headers)
+            start_event.set()
+
             await self._reset_timeout()
             async for msg in self._ws_connection:
                 if not self._process_message(msg):
                     break
                 await self._reset_timeout()
+        except Exception as e:  # pylint: disable=broad-except
```
bdraco, line 85: Can you make this less broad ?

AngellusMortis: Networking errors tend to be very broad, and it is hard to catch them all. I could make it the aiohttp base exception, but it maybe miss others. There is an `except Exception` further up the line that catches anything that was missed, but this does ensure the errors are always logged properly for us.

bdraco: The one a few lines up seems fine since it is protecting against code we don't control.  We should be able to know all the exceptions the downline will throw in this case.

bdraco (approved):

PR #166, opened 2022-01-24
Adds doorlock class

AngellusMortis:
Initial Doorlock model. I plan to do a follow up to change the auto-lock settings, locking/unlocking and the LED settings.

Exclude test data: https://github.com/briis/pyunifiprotect/pull/166/files?file-filters%5B%5D=.py


pyunifiprotect/data/types.py
```diff
@@ -256,6 +256,18 @@ class ChimeType(int, enum.Enum):
...
+@enum.unique
+class LockStatusType(str, ValuesEnumMixin, enum.Enum):
+    OPENING = "OPENING"
+    CLOSING = "CLOSING"
+    OPEN = "OPEN"
+    CLOSED = "CLOSED"
+    NOT_CALIBRATED = "NOT_CALIBRATED"
+    AUTO_CALIBRATION_IN_PROGRESS = "AUTO_CALIBRATION_IN_PROGRESS"
+    CALIBRATION_WAITING_CLOSE = "CALIBRATION_WAITING_CLOSE"
+    CALIBRATION_WAITING_OPEN = "CALIBRATION_WAITING_OPEN"
```
bdraco, line 268: Is there no jammed status?

bdraco (approved):

pyunifiprotect/data/types.py
AngellusMortis: There is. I pulled those from the JS. let me dig again and see if I can find the jammed status. Will post it in the second PR.

PR #170, opened 2022-05-06
Require newer version of typer to ensure click compatbility

AngellusMortis:
typer < 0.4.1 is not compatible with click >= 8.1.0. However, 0.4.1 should be compatible with both click < 8.10 and click >= 8.1.0.

Fixes https://github.com/home-assistant/core/issues/71422

### DIFF
--- a/pyunifiprotect/utils.py
+++ b/pyunifiprotect/utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
-from ast import Import
 
+from ast import Import
 import asyncio
 from collections import Counter
 import contextlib
@@ -38,13 +38,6 @@ from pydantic.utils import to_camel
 
 from pyunifiprotect.data.types import Version
 from pyunifiprotect.exceptions import NvrError
-
-# typer<0.4.1 is incompatible with click>=8.1.0
-# allows only the CLI interface to break if both are installed
-try:
-    import typer
-except ImportError:
-    typer = None  # type: ignore
 
 if TYPE_CHECKING:
     from pyunifiprotect.api import ProtectApiClient
@@ -305,6 +298,10 @@ async def write_json(output_path: Path, data: Union[List[Any], Dict[str, Any]]) 
 
 
 def print_ws_stat_summary(stats: List[WSStat], output: Optional[Callable[[Any], Any]] = None) -> None:
+    # typer<0.4.1 is incompatible with click>=8.1.0
+    # allows only the CLI interface to break if both are installed
+    import typer  # pylint: disable=import-outside-toplevel
+
     if output is None:
         if typer is not None:
             output = typer.echo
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     pyjwt
     python-dotenv
     pytz
-    typer
+    typer>=0.4.1
```
bdraco, line 33: Ideally we defer loading this with a late import when using library headless.

bdraco (approved):

setup.cfg
AngellusMortis: I actually do not know why it is being imported as it is _not_ being imported anywhere in the core lib, only the CLI interface. I am guessing there is a something with how Python automatically imports `__main__.py` when loading `__init__.py` that causes it to be imported.

bdraco:
Maybe drop `__main__.py`, move `start` into `cli.py`, and change setup.cfg to something like?
```
[options.entry_points]
console_scripts =
        unifiprotect = pyunifiprotect.cli:start
```

bdraco (approved):

setup.cfg
AngellusMortis: Tracked down the actual cause for it being imported early and fixed it. Also, removing the `__main__.py` would mean you cannot do `python -m pyunifiprotect` instead of `unifi-protect` (in case the script wrapper did not get installed correctly or something)

pyunifiprotect/utils.py
```diff
@@ -34,11 +35,17 @@
 from aiohttp import ClientResponse
 from pydantic.fields import SHAPE_DICT, SHAPE_LIST, ModelField
 from pydantic.utils import to_camel
-import typer
 
 from pyunifiprotect.data.types import Version
 from pyunifiprotect.exceptions import NvrError
 
+# typer<0.4.1 is imcompatible with click>=8.1.0
```
bdraco, line 42:
```suggestion
# typer<0.4.1 is incompatible with click>=8.1.0
```

pyunifiprotect/utils.py
```diff
@@ -34,11 +35,17 @@
...
+# allows only the CLI interface to break if both are installed
+try:
```
bdraco, line 44: Even better if you import it in print_ws_stat_summary

AngellusMortis: I thought about that but figured this was better since I was not importing inside of a function. I guess it is either ignoring mypy error or ignoring pylint error. lol.

Does not matter either way really in the end since it would break anyways if it imported with the bad versions.

bdraco: Ignoring the pylint for the late import seems completely reasonable in this case

bdraco (approved):
🚀

PR #172, opened 2022-05-19
Adds play_buzzer for chime. Breaks up test_data tests

AngellusMortis:
* Adds `play_buzzer` for Chimes
* Breaks up `test_data` tests into multiple files
    * Because this is a lot of changes, the only _actually_ changed test has been marked
* Adds missing classes from `.data.__init__`.


tests/data/test_chime.py
```diff
@@ -0,0 +1,182 @@
...
+async def test_chime_play_buzzer(chime_obj: Optional[Chime]):
+    if chime_obj is None:
+        pytest.skip("No chime_obj obj found")
+
+    await chime_obj.play_buzzer()
+
+    chime_obj.api.api_request.assert_called_with(
+        f"chimes/{chime_obj.id}/play-buzzer",
+        method="post",
+    )
```
AngellusMortis, line 66: This is the only test that is actually changed/added/removed. 

All other tests were just moved into other files to break up the massive file `test_data_updates.py`

bdraco: Thanks. That helps  👍

bdraco (approved):

PR #174, opened 2022-05-28
Various project updates

AngellusMortis:
* Removes tox and add helper scripts to run format/lint/test and updates GHA workflows to use them
* Switch devcontainer to use python user instead of root
    * This allows VS Code to re-map the UID/GIDs to prevent permissions issues
* Updates VS Code devcontainer config to use newer python extensions
* Removes old VS Code tasks
* Updates black and pytest configs
* Removes old workspace (not really needed for only a single folder)
* Removes test CLI command
* Updates requirements and simplifies them to a single file for CI
    * pydantic was temporarily pinned to 1.9.0 (did not want to make code changes for this PR)


bdraco (approved):

bdraco:
Looks good once the deploy failure is fixed unless its unrelated

AngellusMortis:
It is unrelated. I cannot do the live data tests/deploys from other branches. It only works from master.

PR #175, opened 2022-05-29
Adds NVR CLI interface

AngellusMortis:
Start of adding a full CLI interface for the library.

* Moves shared CLI options to `unifi-protect` root command
    * Technically a breaking change, but the CLI interface before this PR was only for CI and debugging. This is the start of stabilizing it. The preferred way is still via ENVs anyways.
* Refactors the root commands to use a context obj to initialize and share UniFi Protect instance
* Adds `nvr` command group that provides interfaces for interacting with `NVR` class (works great with `jq`)
* Fixes child `ProtectBaseObj` not being serialized correctly for `save_device`

### DIFF
--- a/pyunifiprotect/data/nvr.py
+++ b/pyunifiprotect/data/nvr.py
@@ -503,10 +503,6 @@ class DoorbellSettings(ProtectBaseObject):
     def _get_unifi_remaps(cls) -> Dict[str, str]:
         return {**super()._get_unifi_remaps(), "defaultMessageResetTimeoutMs": "defaultMessageResetTimeout"}
 
-    def unifi_dict(self, data: Optional[Dict[str, Any]] = None, exclude: Optional[Set[str]] = None) -> Dict[str, Any]:
-        data = super().unifi_dict(data, exclude)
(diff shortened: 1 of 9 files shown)


pyunifiprotect/data/nvr.py
```diff
@@ -503,6 +503,10 @@ class DoorbellSettings(ProtectBaseObject):
     def _get_unifi_remaps(cls) -> Dict[str, str]:
         return {**super()._get_unifi_remaps(), "defaultMessageResetTimeoutMs": "defaultMessageResetTimeout"}
 
+    def unifi_dict(self, data: Optional[Dict[str, Any]] = None, exclude: Optional[Set[str]] = None) -> Dict[str, Any]:
+        data = super().unifi_dict(data, exclude)
+        return data
+
```
bdraco, line 509: Is this needed ?  It looks like a useless delegation

AngellusMortis: Nice catch. That was for testing.

bdraco (approved):

PR #178, opened 2022-06-04
Adds second round of commands for UFP devices

AngellusMortis:
* Adds `liveview` subcommand
* Adds validated to ensure device/devices are set correct on the context
* Fills out subcommands for all of the helper methods for `chime`, `doorlock`, `light` and `viewer` objects (still missing are sensor and camera)

### DIFF
--- a/pyunifiprotect/cli/base.py
+++ b/pyunifiprotect/cli/base.py
@@ -2,7 +2,7 @@ import asyncio
 from dataclasses import dataclass
 from enum import Enum
 import json
-from typing import Any, Callable, Coroutine, Mapping
+from typing import Any, Callable, Coroutine, Mapping, Sequence
 
 import typer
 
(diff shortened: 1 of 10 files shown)


pyunifiprotect/cli/base.py
```diff
@@ -36,10 +36,23 @@ def json_output(obj: Any) -> None:
...
+    elif output_format == OutputFormatEnum.JSON:
+        json_output(None)
+
+
+def print_unifi_list(objs: Sequence[ProtectBaseObject]) -> None:
+    """Helper method to print a list of protect objects"""
+
+    data = []
+    for obj in objs:
+        data.append(obj.unifi_dict())
```
bdraco, line 53: This could be a comprehension

bdraco (approved):

PR #192, opened 2022-06-13
Adds permission checking to various methods

AngellusMortis:
* Adds `BadRequest` to making API requests
* Adds `revert_changes` method to base ProtectModel, will revert any unsaved changes to the model
* Updates `save_device` to automatically revert any changes made to the device if there was an error saving
* Adds permissions checks to various applicable places
    * Not a breaking change these without this check, it would have raised a `NotAuthorized` error on trying to make the network request. The error has just been moved up to before the network request.

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -264,6 +264,8 @@ class BaseApiClient:
                 if raise_exception:
                     if response.status in (401, 403):
                         raise NotAuthorized(msg % (url, response.status, reason))
+                    if response.status >= 400 and response.status < 500:
+                        raise BadRequest(msg % (url, response.status, reason))
                     raise NvrError(msg % (url, response.status, reason))
                 _LOGGER.debug(msg, url, response.status, reason)
                 return None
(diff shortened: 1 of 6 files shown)


pyunifiprotect/api.py
```diff
@@ -264,6 +264,8 @@ async def api_request_raw(
                 if raise_exception:
                     if response.status in (401, 403):
                         raise NotAuthorized(msg % (url, response.status, reason))
+                    if response.status >= 400 and response.status < 500:
+                        raise BadRequest(msg % (url, response.status, reason))
```
bdraco, line 268: This is a breaking change right?

bdraco (approved):

pyunifiprotect/api.py
AngellusMortis: It is raises BadRequest instead of NVR Error, but both of them inherit from ClientError.

bdraco: I don't think there are many other users of this api so it probably doesn't matter either way

bdraco: famous last words ^

PR #200, opened 2022-06-21
Fix permissions issue

AngellusMortis:
* Adds support for a permission string that has multiple obj IDs
* Adds support for the "self" permission identifier (`$`) for user permissions


bdraco (approved):
Confirmed fixed.  Thanks 👍

PR #203, opened 2022-06-22
Speed up the is_authenticated check

bdraco:
- We can cache the lookup of the token and only look it up again when we know its invalid


bdraco:
Its a lot better
Screen Shot 2022-06-22 at 16 54 30

AngellusMortis (approved):

PR #206, opened 2022-06-23
Improve WS message processing

AngellusMortis:
* Ignore update packets for devices that do not exist
* Catch validation errors for WS packets and trigger a manual refresh of the device in the bootstrap (or ignore them for non-adoptable devices)
* Adds warning log message for using an EA version of UniFi Protect in Home Assistant

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -57,6 +57,15 @@ NEVER_RAN = -1000
 DEVICE_UPDATE_INTERVAL = 900
 # retry timeout for thumbnails/heatmaps
 RETRY_TIMEOUT = 10
+EA_WARNING = """
+You are running an Early Access version of UniFi Protect. Early Access versions of
+UniFi Protect are not supported for Home Asssitant. If you are using Home Asssitant
+and have an error, please report errors to https://github.com/briis/pyunifiprotect
+first. DO NOT REPORT EA ISSUES TO HOME ASSISTANT CORE.
+
+It is recommended you downgrade to a stable version.
+https://www.home-assistant.io/integrations/unifiprotect#downgrading-unifi-protect.
+"""
 
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


pyunifiprotect/data/bootstrap.py
```diff
@@ -452,15 +457,53 @@ def process_ws_packet(
...
+            try:
+                model_type = ModelType(action["modelKey"])
+                device_id = action["id"]
+                asyncio.create_task(self.refresh_device(model_type, device_id))
+            except (ValueError, IndexError):
+                msg = f"{action['action']} packet caused invalid state. Unable to refresh device."
+            else:
+                msg = f"{action['action']} packet caused invalid state. Refreshing device: {model_type} {device_id}"
+
+            _LOGGER.warning("%s Error: %s", msg, err)
```
bdraco, line 490: Can we warn only if it is still invalid after the refresh?

AngellusMortis: I have mixed feeling. If they are running an unsupported version of Protect, this could happen a lot. But if the refresh causes an issue and they report it, without this log message we would never know it could have been from a refresh.

AngellusMortis: No. Nice catch.

Also, how about adding another warning log on creating the API client if the client has not been updated to support that version of Protect yet.

bdraco: > I have mixed feeling. If they are running an unsupported version of Protect, this could happen a lot. But if the refresh causes an issue and they report it, without this log message we would never know it could have been from a refresh.

We could include that we tried to refresh but it still got a bad value in the log message.

> Also, how about adding another warning log on creating the API client if the client has not been updated to support that version of Protect yet? 

Seems like a good idea to me.

AngellusMortis: Added a message that automatically appears if you are running an EA version of UniFi Protect.

pyunifiprotect/api.py
```diff
@@ -56,6 +57,20 @@
...
+EA_WARNING = """
+You are running version %s of UniFi Protect, which is an Early Access version. Early Access versions of UniFi Protect are not supported for Home Asssitant.
+
+If you are using Home Asssitant and have an error, please report errors to https://github.com/briis/pyunifiprotect first. DO NOT REPORT EA ISSUES TO HOME ASSISTANT CORE.
+
+It is recommended you downgrade to a stable version. https://www.home-assistant.io/integrations/unifiprotect#downgrading-unifi-protect.
+"""
+
+try:
+    import homeasssitant  # type: ignore # pylint: disable=unused-import # noqa
```
bdraco, line 69: Maybe check if its in sys.modules instead ?

PR #209, opened 2022-07-04
Use info

ludeeus:
Use info instead of warning.

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -562,7 +562,7 @@ class ProtectApiClient(BaseApiClient):
             bootstrap_updated = True
             self._bootstrap = await self.get_bootstrap()
             if self._last_update == NEVER_RAN and IS_HA and self._bootstrap.nvr.version.is_prerelease:
-                _LOGGER.warning(EA_WARNING, self._bootstrap.nvr.version)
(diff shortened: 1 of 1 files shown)


bdraco:
This is a difficult one since we get a lot of core issues complaining that UFP is broken when users are running early access.  The origin of this message is partly out of frustration with overly demanding users wanting to know when core is going to be fixed.  I'm concerned downgrading the severity will cause more users to miss it, and this is a case where we really want to set expectation up front to avoid the demands.

ludeeus:
> The origin of this message is partly out of frustration

Well, isn't that funny.
This PR exists only out of frustration, this is IMO a bad design, and I have removed the integration from my setup.

AngellusMortis:
The warning is to ensure the most stable experience for users who do not actually understand how Early Access works. We do not have control over Ubiquiti's update schedule, but they thankfully do have Early Access, which gives us the time to update and fix any breakage so there are normal users who stay on stable should _never_ have any breakage. 

If you are able to provide a better way to accomplish that, I am certainly more than willing to hear it, but the other alternative I thought of was "whitelisting" versions of Protect that are "supported", but that creates a much larger amount of work on our part because we would need a version bump _every_ time Protect came out with a new version and add the new version.

PR #212, opened 2022-07-14
Use pydantic dict fast path for update_from_dict

bdraco:
https://github.com/samuelcolvin/pydantic/blob/228e46de35910d5caed504c48b71900fef1d2b8a/pydantic/main.py#L821

Similar to https://github.com/briis/pyunifiprotect/pull/198

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -492,7 +492,12 @@ class ProtectBaseObject(BaseModel):
         for key in data:
             setattr(self, key, convert_unifi_data(data[key], self.__fields__[key]))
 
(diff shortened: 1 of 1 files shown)


bdraco:
There is another one in `construct` but its called far less frequently

```python
        obj = super().construct(_fields_set=_fields_set, **values)
        obj._initial_data = obj.dict(exclude=cls._get_excluded_changed_fields())  # pylint: disable=protected-access
        obj._api = api  # pylint: disable=protected-access
```

Didn't notice it before since the other one isn't hitting the top of the profile anymore its enough that it now shows up but still pretty low

PR #213, opened 2022-08-04
Update README for MFA limitation

davidhariri:


README.md
```diff
@@ -29,6 +29,8 @@ commands support JSON output so it works great with `jq` for complex scripting.
 Following traditional [twelve factor app design](https://12factor.net/), the perfered way to provided authentication
 credentials to provided environment variables, but CLI args are also supported.
 
+Please note that MFA codes are not supported by this library.
```
AngellusMortis, line 32:
```suggestion
Ubiquiti SSO accounts are not supported and actively discouraged from being used. There is no option to use MFA. You are expected to use local access user.
```

davidhariri: Happy to change the copy, but this has nothing to do with SSO accounts. This is for accounts with multi factor authentication enabled.

AngellusMortis: You should not have an SSO account with MFA enabled, if you do, your account is vulnerable, especially if you have remote access on your console enabled. 

In the Home Assistant docs we actually tell the users not to use them.

davidhariri: Again, my PR is not addressing anything to do with SSO accounts. It's for **normal** UniFi user accounts with MFA code sign in enabled (using Google Authenticator, 1Password etc..). This library doesn't work with MFA, so I'm just adding a doc string for that as it wasn't clear .

davidhariri: By SSO account, do you mean a UniFi account?

AngellusMortis: Ubiquiti/UI SSO account is the official name for the accounts: 

image

The intent with the suggestion is allow the PR to still be merged and you get contributor credit. I can make the commit without the PR and close it, but then you do not get contributor credit, but we do not want to encourage users to use SSO accounts as that does require you to have an SSO account with Remote Access enabled and _without_ MFA, which is not secure.

davidhariri: Ah, I understand now, thank you. I didn't know that SSO referred to UniFi SSO (I was thinking 3P SSO like Okta etc). Will update.

PR #217, opened 2022-08-15
Adds initial backup command

AngellusMortis:
Initial command to let you backup events from a camera. Inspired by the project @ep1cman's [unifi-protect-backup](https://github.com/ep1cman/unifi-protect-backup).

The main difference with this version is it does not only get events going forward, but it also exports **all** events that UniFi Protect knows about. It also creates an export of the events into a sqlite database.

For testing, it is recommended to provide a `-s` / `--start` and/or a `-t` value to filter out the events that are exported from UniFi Protect to speed up the construction of the sqlite database that is built. Example:

Backup all smart detections starting on August 15th
```
unifi-protect backup -s 2022-8-15 events -t smartDetectZone
```

This is just an initial version and considered unstable. Still planned:

* Command to backup all video footage from a specific range
* Command to run a long live process that listens for events as they come up and exports video in real-time (basically what unifi-protect-backup does). Watch process will _only_ handle events since trying to back up a full NVR instance is real-time is a larger issue (you need both more space and faster disks then your NVR, which is likely hard to come by).
* Handle camera renames for detecting if files already exist on disk
* Ignoring/selecting cameras
* Documentation for usage

There are no plans for any of the rclone features that unifi-protect-backup has as that is out of scope. rclone should be ran by an external service, like as part of a full system backup. You should have [two backups anyway](https://www.backblaze.com/blog/the-3-2-1-backup-strategy/).


pyunifiprotect/cli/backup.py
```diff
@@ -0,0 +1,491 @@
...
+        result = await db.execute(query)
+        smart_types_set = {s.value for s in smart_types}
+        for event in result.unique().scalars():
+            event = cast(Event, event)
+            if event.event_type == d.EventType.SMART_DETECT:
+                if not event.smart_types.intersection(smart_types_set):
+                    continue
+
+            while sem.locked():
+                await asyncio.sleep(0.1)
```
bdraco, line 436: I think you could try to acquire here, and then release instead of sleep so it wakes up as soon as a free slot is available

pyunifiprotect/cli/backup.py
```diff
@@ -0,0 +1,491 @@
...
+
+            loop = asyncio.get_running_loop()
```
bdraco, line 438: I'd move this outside the loop

pyunifiprotect/cli/backup.py
AngellusMortis: That is actually probably a good idea. So just acquire and then release right away?

pyunifiprotect/cli/backup.py
```diff
@@ -0,0 +1,510 @@
...
+        for event in result.unique().scalars():
+            event = cast(Event, event)
+            if event.event_type == d.EventType.SMART_DETECT:
+                if not event.smart_types.intersection(smart_types_set):
+                    continue
+
+            # wait for a free download slot
+            await sem.acquire()
+            sem.release()
+            loop.create_task(_download_event(ctx, event, force, pb, sem))
```
bdraco, line 458: Untested, but you might be able to do `async with sem` here to avoid writing out the aquire/release

bdraco: To avoid the sleep, you could push them into an array and `await` them all with `asyncio.wait` at the end but that might be a lot of tasks.

AngellusMortis: I would need a `pass` inside of the context. I think it is better to be explicit here.

AngellusMortis: And yeah, the idea is to _not_ load all of the tasks into memory since that could easily be ten of thousands of tasks. Something else I also need to look into for follow up is verify it is using a cursor and only one event if pulled from the database at a time.

AngellusMortis:
unifi-protect-backup is flawed in many ways as it only listens to events from the Websocket. So, it cannot retrieve events before it runs, and it can miss events if a Websocket message is missed for some reason. It would require a decent rewrite to fix that. 

Also, if not noticed by the library being moved to my repo, I am essentially the maintainer for UniFi Protect. Backing up events is only the first part of expanding the CLI that I want to do since it is something users very commonly ask for. There are also APIs to backup and restore settings for Protect itself that would be good to integrate into this.

ep1cman:
Ah my bad, I knew the name sounded familiar!

bdraco (approved):

AngellusMortis:
Going to start work on improving the `events` subcommand. I have all of my events dumpped out of my instance, but now I want to better handle the errors while downloading the files, verifying the files on disk, handle camera renames and tag the files with metadata.

PR #218, opened 2022-08-16
Updates for AI Theta

AngellusMortis:
* Adds `hotplug` and `lens_type` for feature flag
* Adds helper to return bool if video playback is possible since AI Theta can have its camera lens removed.


bdraco (approved):

bdraco:
Tests need a little work, but you probably know that

AngellusMortis:
I think it is time for me to generate new test data, but that takes effort to get all of the events and data in to it.

PR #223, opened 2022-08-26
Updates for Protect 2.2.1-beta5

AngellusMortis:
Updates from latest beta:

* Adds `setInsights` helper for NVR since the value is now exposed in the Web UI
* Adds groundwork for audio smart detections
* Adds properties to help detect if a camera has a removable speaker and whether it is attached (new AI Theta audio module)


bdraco (approved):
LGTM 👍

PR #225, opened 2022-08-31
Updates pinned reqs for testing

AngellusMortis:


bdraco (approved):
Looks fine.  At EWR dealing with a connection snafu due to the thunderstorms so can’t test locally though

PR #226, opened 2022-08-31
Aggregate updates to reduce PATCH requests/locking

AngellusMortis:
* Adds `queue_update` method that pushes an update onto a queue and then waits for others updates to come in (50 ms). Will automatically cancel if one does. Only the last remaining `queue_update` will then acquire the update lock and call `save_device`
* Updates all of the helper methods to use the new `queue_update` method


bdraco (approved):
LGTM, but I'm traveling so I can't properly test

AngellusMortis:
From my testing, it looks like it fixes the original issue.

PR #229, opened 2022-09-09
Adds relative datetimes for backup start/end

AngellusMortis:
* Adds `dateparser` requirement
* Allows human readable/relative datetime parsing from command line for start/end on backup events command (really useful if you want to put it on a cron)


bdraco (approved):
FYI, dataparser is relatively slow.

It doesn't matter here though

PR #231, opened 2022-09-10
Adds file verification for backup

AngellusMortis:
* Handle camera renames for files on disk
* Adds verification of files on disk using ffmpeg/pillow. Since this takes significantly longer to do, it is a flag that must be passed to complete it

Had to tweak the threshold for what is consider a "valid" video export since they vary so much. It looks 80% (18 seconds of a 22 second event) of the original event length seems to be the best. From sampling my events on my UNVR Pro, any event under 80% length has parts of the video missing.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ install_requires =
     aiohttp
     aioshutil
     async-timeout
+    av
     dateparser
     orjson
     packaging
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,6 @@ install_requires =
     aiohttp
     aioshutil
     async-timeout
-    av
     dateparser
     orjson
     packaging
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     aiohttp
     aioshutil
     async-timeout
+    av
```
bdraco, line 30: It would be good to load this only in the cli since hass uses `ha-av` when `av` gets too far behind and they don't play well together

https://pypi.org/project/ha-av/

AngellusMortis: Oh. I thought HA got rid of the fork. I guess it is still around. Moved it to the `backup` extra.

bdraco (approved):

PR #232, opened 2022-09-10
Adds metadata to backups and more file context

AngellusMortis:
* Adds new title value for the file context that is used for writing video metadata by default
* Adds metadata to video files, metadata includes creation time/year/release and the a human friendly title for the file. Title can be controlled by the new `--title-format` parameter. Defaults to `"{time_pretty_local} {sep} {camera_name} {sep} {event_type_pretty} {sep} {length_pretty}"`
* Adds plenty of new fields for file contexts for setting templates for title/output files, including a lot of `_local` fields so you can save files in the same timezone as the camera
* Runs verification/metadata operations in a separate thread
* Removes pytz as a main dep since `zoneinfo` is python 3.9+

Note: ideal format for Plex to ingest exported files is:
```
--thumb-format "{year_local}/{month_local}/{day_local}/{hour_local}/{title}.jpg" --event-format "{year_local}/{month_local}/{day_local}/{hour_local}/{title}.mp4"
```

### DIFF
--- a/pyunifiprotect/utils.py
+++ b/pyunifiprotect/utils.py
@@ -481,7 +481,7 @@ def get_local_timezone() -> tzinfo:
         return TIMEZONE_GLOBAL
 
     try:
-        from homeassistant.util import (  # pylint: disable=import-outside-toplevel  # type: ignore
+        from homeassistant.util import (  # type: ignore  # pylint: disable=import-outside-toplevel
             dt as dt_util,
         )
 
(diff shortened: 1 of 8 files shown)


bdraco (approved):

AngellusMortis:
Marking draft since I already ran into some errors running it against my NVR.

bdraco (approved):

pyunifiprotect/utils.py
```diff
@@ -439,3 +442,81 @@ def decode_token_cookie(token_cookie: Morsel[str]) -> Dict[str, Any] | None:
...
+    except ImportError:
+        pass
+
+    timezone_name = os.environ.get("TZ")
+    if timezone_name:
+        return _set_timezone(timezone_name)
+
+    timezone_name = "UTC"
+    timezone_locale = Path("/etc/localtime")
+    if timezone_locale.exists():
```
bdraco, line 498: Just a note:

This does I/O, and I didn't see any async usage of this.  If you are planning on using it in hass, it might be nice to have an async wrapper that runs it in the executor

AngellusMortis: It will not get that far because of the lines above. Which is a special case for HA.

PR #243, opened 2022-10-17
Remove out of place warning log

frenck:
This PR removes a warning log, that should not be a responsibility of a library.

In this case, this library tries to detect Home Assistant and log a warning in case a certain version is out of bounds. As libraries should be agnostic, this is out of bounds.

If Home Assistant would need such a warning, it should be implemented on the integration (and not the underlying library).

This PR, therefore, removes the warning log from this package and cleans up the constants that were solely involved with/used for this warning log.

../Frenck

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -60,15 +60,6 @@ NEVER_RAN = -1000
 DEVICE_UPDATE_INTERVAL = 900
 # retry timeout for thumbnails/heatmaps
 RETRY_TIMEOUT = 10
-EA_WARNING = """
-You are running version %s of UniFi Protect, which is an Early Access version. Early Access versions of UniFi Protect are not supported for Home Asssitant.
-
-If have an error, please report errors to https://github.com/AngellusMortis/pyunifiprotect first. DO NOT REPORT EA ISSUES TO HOME ASSISTANT CORE.
-
-It is recommended you downgrade to a stable version. https://www.home-assistant.io/integrations/unifiprotect#downgrading-unifi-protect.
-"""
-IS_HA = "homeassistant" in sys.modules
-
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -598,9 +589,6 @@ class ProtectApiClient(BaseApiClient):
         if self._bootstrap is None or now - self._last_update > DEVICE_UPDATE_INTERVAL:
             bootstrap_updated = True
             self._bootstrap = await self.get_bootstrap()
-            if self._last_update == NEVER_RAN and IS_HA and self._bootstrap.nvr.version.is_prerelease:
-                _LOGGER.warning(EA_WARNING, self._bootstrap.nvr.version)
-
             self._last_update = now
             self._last_update_dt = now_dt
 
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -7,7 +7,6 @@ from http.cookies import Morsel
 from ipaddress import IPv4Address
 import logging
 from pathlib import Path
-import sys
 import time
 from typing import Any, Callable, Dict, List, Optional, Set, Type, Union, cast
 from urllib.parse import urljoin


AngellusMortis:
Duplicate PR: https://github.com/AngellusMortis/pyunifiprotect/pull/209

> As libraries should be agnostic, this is out of bounds.

First, this is my repo. The library can be however we want it to be. We put this here in the pyufp library because we believe it would not be accepted into core. This repo is not in any way associated or controlled by Home Assistant.

We made the decision very seriously to put this in this repo instead because **we do not want users to try to suppress it or remove it**. If this log message were added to core, it could very easily be later removed the first time a core developer is annoyed by the message. We _want_ the message there as UniFi Protect does not have an official API. Upgrading to EA versions can and _will_ break your Home Assistant install. Users need to aware of the risk of doing so. If you do not want the log, do not use EA versions of Protect.

In the event a user _chooses_ to install an EA release, we want to redirect support issues to this repo instead of Home Assistant as they will 9 times out of 10 be a pyufp issue due to a breaking API change.

AngellusMortis:
EA versions not being official supported also gives me more time to update pyufp after an EA release comes out. I _usually_ have the updates completed with 12 hours of the release so they will be in the next HA minor release, but there is no promise of that because of life and all of that. 

So, EA versions much stay "unsupported" at all times to ensure I have enough time to fix anything that breaks before it goes to stable and everyone auto-updates to the version.

AngellusMortis:
Unfortunately, HA development standards documentation is also very lax as well and I have run into multiple cases were different core devs gave different conflicting "Requests Changes". HA core is also not my project either, so I have no control over what core devs decide to do with code I wrote, _but_ I am still expected to be the code owner of said code and support it in the event issues occur.

As a result, for this PR to be accepted, I would need to see the following from the core side for HA:

1. An official developers' documentation on how to handle supported versions of upstream APIs. If this is as simple as "add a warning log message if the user is using an unsupported upstream API", but it needs to be in writing with a git history log. That means it trackable and transparent and something that can be referenced later. 
     * It could be something as nice as a log message/persistent notification + some config entry option that allows the user to accept the warning and suppress it, but for the log to be suppressed, there needs to an action from the user that shows in a diagnostic saying "I was told this API was unsupported and choose to ignore it". 
2. The result of 1 implemented in core, but a core dev. So again, in git history as to why this was added.

Both of these would give me the confidence that if warning was added to core it would not later be removed by the whim of a core dev.

frenck:
Fine, will upgrade upstream with a fork of this library instead. 🤷

frenck:
@AngellusMortis This library is forked at the point is is listed as MIT: <https://github.com/AngellusMortis/pyunifiprotect/blob/master/LICENSE>

Feel free to change the license at this point.

AngellusMortis:
The reason was stated very plainly:

> we do not want users to try to suppress it or remove it.

> HA core is also not my project either, so I have no control over what core devs decide to do with code I wrote

I have **zero** confidence that is this log message was added to core without some kind of documentation or policy it is allowed, the first core dev that saw it would remove it. Because after all, me contributing my time to HA core as code owner of an integration cleanly means nothing core developers.

AngellusMortis:
Code owners of integrations have zero voice in how HA core is run and operated. It is an "open source project ran by the community", but in reality a very small handful of users (the core devs) have absolutely control over the project. 

I _as an integration code owner_ do not want my code to be used or my time to be put towards a project where I cannot provide adequate warning to users about the risk of running unsupported APIs against my code. All because another dev disagrees about whether or not a log message is unhelpful.

AngellusMortis:
By telling me "I will fork this project and do it anyways and add it to core" reinforces the point I am making as well. Core devs do not see code owners of integrations as have any voice in how HA runs. 

**_I_** support the `unifiprotect` integration with my time and effort. I add a simple log message that states using EA versions of Protect can break HA (because it is true) and now because a core dev disagrees with the usefulness of that message, that message will be erased, and I will still be expected to support the code without my warning place.

frenck:
> How do you not see the issue in that? 

I never made any statement on that? As a matter of fact, I only made this change because of an area of responsibility mismatch and from a software design perspective (as per the PR description).

Instead of a healthy question back (as a response to this PR) on how to resolve such a concern/issue you have, I'm greeted (with what feels like) a wall of hate toward the HA project and its policies, and maybe even personal? (I dunno).

Not sure how you expected me to respond to any of that or why that is.

I still think, from a software design perspective, this PR makes sense (moving it), and I'm happy to discuss it if you have doubts or questions about the suggestion made in this PR.

But as I said, it seems like you have completely different issues with, whatever, that I was not here for at all, nor do I have any interest in getting into those things.

I was just here for an improvement, made in my spare free time, at 1 AM in the night for the love of open source and software design. I'm not even using this for my personal setup 🤷 

However, your response is worrying for the end-users of Home Assistant, and thus I made a fork to ensure its existence in the current state and its current license; just in case.

Feel free to get back to me, if you feel like that is the right thing to do. I'm available on Discord any time (well, except right now, this very instance, as it is past 2 AM and I really need to get some sleep 💤).

../Frenck

PR #244, opened 2022-11-06
Adds more accurate detection of prerelease versions

AngellusMortis:
* Removes HA EA Warning has it has been added to HA as a Repair
* Adds new API method to fetch all release versions from the UI APT repos
* Adds new CLI command: `unifi-protect release-versions` that will return all release versions from APT repos and cache them to file inside of the package.
* Adds `get_is_prerelease()` method to `Bootstrap` that checks if the current release is a prerelease version (not in list of all release versions)
    * Return true if version is a prerelease version (`-beta` in release) 
    * Checks `/tmp/ufp_cache/release_cache.json` for release versions otherwise checks `pyunifiprotect/release_cache.json`
    * If no match is found, pulls versions from APT repos and caches them to  `/tmp/ufp_cache/release_cache.json`
    * The intent is that if the user is on a release version, it only ever makes the network release a max of once to get the new release versions


pyunifiprotect/data/bootstrap.py
```diff
@@ -508,3 +516,37 @@ async def refresh_device(self, model_type: ModelType, device_id: str) -> None:
...
+
+        # 2.6.14 is an EA version that looks like a release version
+        versions = await self._read_cache_file(TMP_RELEASE_CACHE) or await self._read_cache_file(RELEASE_CACHE)
+        if versions is None or self.nvr.version not in versions:
+            versions = await self.api.get_release_versions()
+            try:
+                _LOGGER.debug("Fetching releases from APT repos...")
+                await aos.makedirs(TMP_RELEASE_CACHE.parent, exist_ok=True)
+                async with aiofiles.open(TMP_RELEASE_CACHE, "wb") as cache_file:
+                    await cache_file.write(orjson.dumps([str(v) for v in versions]))
```
bdraco, line 548: Should we write to a tempfile and then rename it in place so we never get a partial read of data?

AngellusMortis: Probably a good idea.

bdraco (approved):

PR #258, opened 2023-01-09
Use Recording Mode for motion/smart status

AngellusMortis:
From discussion on this [HA issue](https://github.com/home-assistant/core/issues/83912) and bugging UniFi Protect devs, the intent of setting "Recording Mode" to "Never" is that the camera will not track anything. No motion events and footage. 

Updates all of the "is___detection_on" properties to take this into account and adds an `is_motion_detection_on` property to use for motion.

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -929,12 +929,34 @@ class Camera(ProtectMotionDeviceModel):
         return index is not None
 
     @property
+    def is_recording_enabled(self) -> bool:
+        """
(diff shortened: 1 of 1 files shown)


bdraco (approved):

eluce2:
This behavior is confusing to me. I used to connect my g4 doorbell camera via HomeBridge and was able to get motion events even though I'm never recording. I'm using a UDM Pro without a hard drive installed, so even if I set the recording mode away from "never" it gets reverted. Do I need to get a hard drive just to re-enable motion events from my doorbell camera?

PR #273, opened 2023-04-16
Fix weak refs

AngellusMortis:
Fixes https://github.com/AngellusMortis/pyunifiprotect/issues/267

### DIFF
--- a/pyunifiprotect/data/bootstrap.py
+++ b/pyunifiprotect/data/bootstrap.py
@@ -2,14 +2,13 @@
 from __future__ import annotations
 
 import asyncio
-from collections import deque
 from copy import deepcopy
 from dataclasses import dataclass
 from datetime import datetime
 import logging
 from pathlib import Path
 from tempfile import gettempdir
-from typing import Any, Deque, Dict, List, Optional, Set, Tuple, cast
+from typing import Any, Dict, List, Optional, Set, Tuple, cast
 from uuid import UUID
 
 from aiohttp.client_exceptions import ServerDisconnectedError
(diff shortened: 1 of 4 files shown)


pyunifiprotect/data/bootstrap.py
```diff
@@ -488,7 +490,7 @@ def _handle_ws_error(self, action: Dict[str, Any], err: Exception) -> None:
             try:
                 model_type = ModelType(action["modelKey"])
                 device_id = action["id"]
-                asyncio.create_task(self.refresh_device(model_type, device_id))
+                self._refresh_tasks.append(asyncio.create_task(self.refresh_device(model_type, device_id)))
```
bdraco, line 493: We could make refresh tasks a set and use add_callback_done on the task to discard the task from the set when it's done. The task is passed as the argument to add_callback_done so the callback can be self._refresh_tasks.discard

AngellusMortis: If multiple refreshes happen at once, that could cause a race condition. A task could be queued later, but completed first.

bdraco: I'm probably not explaining that well on my mobile. I'll give you a better example when I'm at by desk

In the mean time. Here is how bleak solved the issue https://github.com/hbldh/bleak/commit/e01ae590df381097df12e9353064627960db4b09

AngellusMortis: Oh, the task is passed back into done callback. Which that makes sense. 

The python docs actually as an example of this: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

bdraco (approved):
tested 👍

PR #274, opened 2023-04-16
Default snapshot to full resolution

AngellusMortis:
UniFi Protect started to make `/snapshot` and `/package-snapshot` default to a height of 360 if no width/height parameters were given. 

This overrides the behavior to always set a `h` query parameter of the max resolution if none are given. 

Fixes https://github.com/AngellusMortis/pyunifiprotect/issues/268

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -1086,19 +1086,19 @@ class Camera(ProtectMotionDeviceModel):
 
     @property
     def high_camera_channel(self) -> Optional[CameraChannel]:
-        if len(self.channels) == 3:
+        if len(self.channels) >= 3:
             return self.channels[0]
         return None
 
     @property
     def medium_camera_channel(self) -> Optional[CameraChannel]:
-        if len(self.channels) == 3:
+        if len(self.channels) >= 3:
             return self.channels[1]
         return None
 
     @property
     def low_camera_channel(self) -> Optional[CameraChannel]:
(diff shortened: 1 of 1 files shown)


bdraco:
LGTM. Will test when I'm back home later today

bdraco:
I'm still getting 360 height.  Doing some more investigation

AngellusMortis:
It would still default to `None` for the height if it cannot locate the high camera channel for some reason.

bdraco:
> It would still default to `None` for the height if it cannot locate the high camera channel for some reason.

Thats probably it.  I have a pretty screwed up test setup right now.  I'm going to test it on production instead

bdraco:
1920x1080 on production 👍

bdraco (approved):

AngellusMortis:
Can you check your other instance and see if it is getting any camera channels? I could just default it to use the first available one instead.

bdraco:
The test instance was running an ancient version of protect. Its broken in so many other ways that its not worth trying to support.

cyberpower678:
This doesn't seem to be fixed.  When specifying the height, it's just scaling up a 360p image it seems like.

AngellusMortis:
That is an issue you have to take up with Ubiquiti. There is no other endpoint in protect that allows you do pull live camera snapshots.

cyberpower678:
Well then, I say we all band together and spam Ubiquiti with support tickets for the protect application, so they are more inclined to do something about it.

PR #277, opened 2023-04-27
Fix missing read only fields exclude

bdraco:
fixes https://github.com/home-assistant/core/issues/92108

### DIFF
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -508,7 +508,9 @@ class ProtectBaseObject(BaseModel):
 
         read_only_fields = self.__class__._get_read_only_fields()  # pylint: disable=protected-access
         excludes = self.__class__._get_excluded_changed_fields()  # pylint: disable=protected-access
-        self._initial_data = {k: v for k, v in self.__dict__.items() if k not in excludes and k not in read_only_fields}
+        self._initial_data = {
+            k: v for k, v in self.__dict__.items() if k == "type" or k not in excludes and k not in read_only_fields
+        }
         return self
 
     def get_changed(self) -> Dict[str, Any]:
--- a/pyunifiprotect/data/base.py
+++ b/pyunifiprotect/data/base.py
@@ -506,11 +506,9 @@ class ProtectBaseObject(BaseModel):
         for key in data:
             setattr(self, key, convert_unifi_data(data[key], self.__fields__[key]))
 
-        read_only_fields = self.__class__._get_read_only_fields()  # pylint: disable=protected-access
         excludes = self.__class__._get_excluded_changed_fields()  # pylint: disable=protected-access
-        self._initial_data = {
-            k: v for k, v in self.__dict__.items() if k == "type" or k not in excludes and k not in read_only_fields
-        }
+        exclude_fields = self.__exclude_fields__
+        self._initial_data = {k: v for k, v in self.__dict__.items() if k in exclude_fields and k not in excludes}
         return self
 
     def get_changed(self) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


bdraco:
I'm out of time to make this work so I'm just going to revert the change

bdraco:
I think this will work and I just hit another existing issue since reverting back to .1 still has the issue with setRecording

bdraco:
Will try to figure out what’s causing that later today. Maybe just a missing exclude

bdraco:
‘’’
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 199, in handle_call_service
    await hass.services.async_call(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1851, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1891, in _execute_service
    await cast(Callable[[ServiceCall], Awaitable[None]], handler.job.target)(
  File "/usr/src/homeassistant/homeassistant/helpers/entity_component.py", line 226, in handle_service
    await service.entity_service_call(
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 798, in entity_service_call
    future.result()  # pop exception if have
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 981, in async_request_call
    await coro
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 838, in _handle_entity_call
    await result
  File "/usr/src/homeassistant/homeassistant/components/unifiprotect/switch.py", line 436, in async_turn_off
    await self.entity_description.ufp_set(self.device, False)
  File "/usr/src/homeassistant/homeassistant/components/unifiprotect/models.py", line 116, in ufp_set
    await getattr(obj, self.ufp_set_method)(value)
  File "/usr/local/lib/python3.10/site-packages/pyunifiprotect/data/devices.py", line 1289, in set_status_light
    await self.queue_update(callback)
  File "/usr/local/lib/python3.10/site-packages/pyunifiprotect/data/base.py", line 642, in queue_update
    await self.save_device()
  File "/usr/local/lib/python3.10/site-packages/pyunifiprotect/data/base.py", line 685, in save_device
    raise BadRequest(f"The following key(s) are read only: {read_only_keys}")
pyunifiprotect.exceptions.BadRequest: The following key(s) are read only: {'isRecording'}
‘’’


only seems to happen if I switch recording mode and than the status light.  It might actually be a race where the previous change had not been sent yet as the problem happens in earlier versions as well

bdraco:
This does fix the reported issue though so might be better to figure the other one out in  another pr

bdraco:
This is taking way more time to test because I’m traveling and can’t do it locally so I may wait to dig into the status light switch issue until I get back home on Saturday.  

I will work on this a bit more later today between meetings

bdraco:
The {'isRecording'} problem only happens if you change it twice quickly.  Probably the change hasn’t taken effect yet which is why it fails.

bdraco:
I double checked that the `{'isRecording'}` issue happens on 4.8.1 so this fix should be fine for the HA issue.

I'll look into trying to tackle the `{'isRecording'}` one this weekend but I don't want to hold up this fix since its not the same problem

bdraco:
To replicate the other issue, change Recording mode from Always to Detections and than quickly change the status light switch.  I double checked its present in 4.8.1.  So this should be good to go

bdraco:
The problem happens when recording turns off and the value of `isRecording` flips and than we try to do a diff and its in there.  So thats an existing issue

bdraco:
Waiting a few minutes and the problem comes back

bdraco:
Its the same issue. Its working the same as 4.8.1 with this change

PR #279, opened 2023-04-27
Revert "Cache static field construction"

bdraco:
Reverts AngellusMortis/pyunifiprotect#275

fixes https://github.com/home-assistant/core/issues/92108


bdraco:
This still doesn't fix the problem so I think something else is going on here.

```Failed to call service select/select_option. The following key(s) are read only: {'isRecording'}```

PR #280, opened 2023-04-27
Fix races when generating diff to save

bdraco:
To fix a race condition with making changes the approach has changed.

1. Once we acquire the update lock, we yield to the event loop to process any incoming websocket messages to sync state before generating the update diff.

2. We no longer use `self._initial_data`, instead we get the state before applying changes, apply changes, than get the new state to generate the diff. We do all of this without yielding via await to the event loop to ensure nothing can change the state between getting the state before changes, making the changes, and diffing the state.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11"
 ]
-version = "4.8.2"
+dynamic = ["version"]
 dependencies = [
     "aiofiles",
     "aiohttp",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11"
 ]
-dynamic = ["version"]
+version = "4.8.2"
 dependencies = [
     "aiofiles",
     "aiohttp",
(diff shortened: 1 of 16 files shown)


bdraco:
There is still a race 

1. Recording mode changes to Detection
2. Other setting is change
3. isRecording changes to False
4. Other setting fails

bdraco:
Its like the isRecording key/value pair is being dropped somewhere

bdraco:
```
2023-04-27 17:13:03.664 DEBUG (MainThread) [pyunifiprotect.data.base] <class 'pyunifiprotect.data.devices.Camera'>: dict diff: {'is_recording': True, 'recording_settings': {'mode': <RecordingMode.ALWAYS: 'always'>}}
2023-04-27 17:13:08.541 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2023-04-27 17:13:11.220 DEBUG (MainThread) [pyunifiprotect.data.base] _generate_update_diff: type=<class 'pyunifiprotect.data.devices.Camera'>, excludes={'last_smart_detect_event_id', 'last_smart_audio_detect_event_id', 'last_smart_audio_detect', 'talkback_stream', 'last_smart_detect', 'last_ring_event_id'}, changed={'isRecording': True, 'ledSettings': {'isEnabled': True}} updated={'isRecording': True, 'ledSettings': {'isEnabled': True}}

```

So it shows changed, we call the switch, and than when we diff again its back to being changed again

bdraco:
trying to debug this over a laptop with a high latency connection on the other side of the world is useless

I'll need to wait till I get back home.  It should go much much faster

bdraco:
20744e22149f16034ef9230dacbda357a9ec4448 is the working SHA

I did too much stripping out

AngellusMortis (approved):

PR #286, opened 2023-05-21
Fix performance regression in `convert_unifi_data` from IPv6 support

bdraco:
convert_unifi_data

### DIFF
--- a/pyunifiprotect/utils.py
+++ b/pyunifiprotect/utils.py
@@ -42,6 +42,7 @@ from pydantic.fields import SHAPE_DICT, SHAPE_LIST, SHAPE_SET, ModelField
 from pydantic.utils import to_camel
 
 from pyunifiprotect.data.types import (
+    Color,
     SmartDetectAudioType,
     SmartDetectObjectType,
     Version,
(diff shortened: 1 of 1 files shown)


pyunifiprotect/utils.py
```diff
@@ -201,33 +204,28 @@ def to_camel_case(name: str) -> str:
 def convert_unifi_data(value: Any, field: ModelField) -> Any:
     """Converts value from UFP data into pydantic field class"""
     from pyunifiprotect.data import (  # pylint: disable=import-outside-toplevel
-        Color,
```
bdraco, line 204: I couldn't figure out why this one was imported outside top level

AngellusMortis: Likely a circular import for something in data that needed something from utils. Or to prevent the possibility of creating one.

bdraco: I changed it to come from `.types` instead since we already already importing there so there shouldn't be a risk of a circular import.  Maybe it used to be somewhere else.

AngellusMortis (approved):

PR #289, opened 2023-06-05
Track every smart detect event seperately

AngellusMortis:
Previously if multiple smart detections were detected at the same time, it would bundle multiple smart detection types into a single event. It looks like newer versions of Protect do the opposite, one event per smart detection type. 

Using the logic to only track the latest ones means that it cannot detect multiple smart detect events if they overlap with each other. This refactors everything to track all of the different types separately.

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -827,6 +827,10 @@ class Camera(ProtectMotionDeviceModel):
     last_smart_audio_detect: Optional[datetime] = None
     last_smart_detect_event_id: Optional[str] = None
     last_smart_audio_detect_event_id: Optional[str] = None
+    last_smart_detects: dict[SmartDetectObjectType, datetime] = {}
+    last_smart_audio_detects: dict[SmartDetectAudioType, datetime] = {}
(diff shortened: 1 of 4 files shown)


pyunifiprotect/data/devices.py
```diff
@@ -946,18 +962,136 @@ def last_ring_event(self) -> Optional[Event]:
...
         return self.api.bootstrap.events.get(self.last_smart_audio_detect_event_id)
 
+    def get_last_smart_audio_detect_event(self, smart_type: SmartDetectAudioType) -> Optional[Event]:
+        """Get the last smart audio detect event for given type."""
+
+        event_id = self.last_smart_audio_detect_event_ids.get(smart_type)
+        if event_id is None:
+            return None
+
+        return self.api.bootstrap.events.get(event_id)
```
bdraco, line 1069: Need to go look and see what cleans these up

bdraco: its a `FixSizeOrderedDict` so no leak here 👍

bdraco (approved):

bdraco:
Tested on production.  No observed unintended side effects

PR #290, opened 2023-06-07
Override lastMotion behavior from unifi protect

AngellusMortis:


bdraco (approved):

bdraco:
confirmed this fixes the state machine behind slammed with updates 👍

PR #292, opened 2023-06-10
Update stream.py to not require AAC encoder

starsoccer:
From a recent HA update, 2023.6 it seems using the speaker of the device no longer works due to a missing AAC encoder as mentioned here, https://github.com/home-assistant/core/issues/94378

Credit to @band8266 for this fix. Ive tested and confirmed this works on my speakers by patching pyunifiprotect

### DIFF
--- a/pyunifiprotect/stream.py
+++ b/pyunifiprotect/stream.py
@@ -132,7 +132,7 @@ class TalkbackStream(FfmpegCommand):
         cmd = (
             "-loglevel info -hide_banner "
             f'{input_args}-i "{content_url}" -vn '
-            f"-acodec {camera.talkback_settings.type_fmt} -ac {camera.talkback_settings.channels} "
+            f"-ac {camera.talkback_settings.channels} "
             f"-ar {camera.talkback_settings.sampling_rate} -b:a {bitrate} -map 0:a "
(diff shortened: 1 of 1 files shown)


AngellusMortis:
Talkback only supports aac.

starsoccer:
🤔 Why would the above patch work for my speakers then? Perhaps I am missing something maybe it is still encoding it as aac?

Sorry I dont have much ffmpeg experience, can just speak from first hand experience that making this change seems to work and not cause the aac encoder issue mentioned.

AngellusMortis:
I do not really care why it works. It may just be dumb luck that it works on your specific camera with that firmware it has installed. The point is that the official UniFi Protect apps only use AAC as confirmed via the Android app, 

The whole point of this library is to provide a Python interface to UniFi Protect and to do things in the same way that the official apps do. Doing anything in a way the official apps do not can cause you to break your camera(s) and potentially void your warranty on your camera(s).

starsoccer:
Noted, thanks for clarifying, did not mean to offend in any way. Ive made an issue in the other repo you mentioned to see about getting it enabled/fixed. 

While I know you have no intent of changing the way this library works just to note in case it is version dependent:
G4 instant: 4.63.22
G4 Dome: 4.63.22
G4 Doorbell Pro: 4.63.22

PR #295, opened 2023-06-21
Always return a string for connection_host

bdraco:
Connection host is only used for generating urls. Each time the URL is generated its having to do strig_from_ip_int under the hood which happens quite frequently when there are a large number of cameras as rtsp_url and rtsps_url are called quite frequently when writing state


bdraco:
This is small but got to start somewhere as ufp updates are still the largest chunk of the event loop run time on my houston system

Screenshot 2023-06-21 at 11 38 05 AM

AngellusMortis:
connection host is a value on the API client, which means it is public. Making it a str instead of an IP address makes the value less versatile and while internally, it only uses it for making URLs, it presents a breaking change for anything else using it.

Maybe an alternative solution is to cache the URLs instead of generating them every time? The only difficulty there is to make sure the cache is blown away if a Websocket update comes in that changes the RTSP alias. I think it should provide better gains as well since even if this is a string, it has to construct it constantly.

bdraco:
> Maybe an alternative solution is to cache the URLs instead of generating them every time? The only difficulty there is to make sure the cache is blown away if a Websocket update comes in that changes the RTSP alias. I think it should provide better gains as well since even if this is a string, it has to construct it constantly.

That works as well.  I wasn't 100% sure which event/update should bust the cache.

AngellusMortis:
The message will [cascade down through the objects](https://github.com/AngellusMortis/pyunifiprotect/blob/master/pyunifiprotect/data/base.py#L469-L478), so you should just need to override the `update_from_dict` method on the camera channel class.

AngellusMortis:
Reading [this code](https://github.com/AngellusMortis/pyunifiprotect/blob/master/pyunifiprotect/data/base.py#L469-L478), you might not need to do anything. I think it would create the channels from scratch if they changed.

bdraco:
It _looks_ like we will get a new channel object so we can cache it as a private attr instead?

bdraco:
closing in favor of https://github.com/AngellusMortis/pyunifiprotect/pull/296

PR #297, opened 2023-07-01
Add pydantic v1 shims to allow pydantic v2

AngellusMortis:
[Pydantic v2 bundled the v1 version with itself](https://github.com/pydantic/pydantic#pydantic-v110-vs-v2), so the first obvious step is to remove the constraint and allow v2 to be installed using the v1 API shim. 

It looks like a lot of the field internals changed so I will have to go back through each import, update them to allow the v2 version and test before we can use v2. It will likely be a breaking change (v5) to get pydantic v2 fully working as much of the "construct" logic may need to be removed or re-written.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,9 +12,13 @@ aioshutil==1.3
     # via pyunifiprotect (pyproject.toml)
 aiosignal==1.3.1
     # via aiohttp
-aiosqlite==0.18.0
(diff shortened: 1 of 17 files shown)


bdraco:
testing now..

bdraco (approved):

bdraco:
nothing blew up

PR #302, opened 2023-08-20
Replace datetime.utcnow

cdce8p:
Starting with Python 3.12 `datetime.utcnow` will be deprecated.
https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow


bdraco (approved):

bdraco:
force pushed to sign the commit

bdraco (approved):

bdraco:
Thanks @cdce8p

PR #318, opened 2023-11-12
[WIP] Added support for Color Night Vision

cmsimike:
Added support to control whether or not Color Night Vision is enabled.

Note: Two tests are currently failing. I tried a few things to get them passing but not sure what I'm missing. If you let me know what I need to do, I'm happy to update this PR.

This addresses: https://github.com/AngellusMortis/pyunifiprotect/issues/303


pyunifiprotect/data/devices.py
```diff
@@ -719,6 +719,7 @@ class CameraFeatureFlags(ProtectBaseObject):
     is_doorbell: bool
     # 2.8.22+
     lens_model: Optional[str] = None
+    has_color_night_vision: Optional[bool] = None
```
AngellusMortis, line 722: You cannot just make up fields that do not exist.

cmsimike: I'm sorry, can you explain a bit more? 
My understanding of the `CameraFeatureFlags` class is that it contains properties which indicate whether or not a particular feature is available for a Protect device. As such I added this property.

cmsimike: Should this align with the data generated by `sample_bootstrap` json of the test data?

AngellusMortis: The classes for fields that _exist in UniFi Protect_. You cannot just arbitrarily add ones.

AngellusMortis:
There still needs to be a boolean or something to determine if a camera has color night vision. It is hard to tell how exactly Protect does that in the Web frontend since it is all obfuscated. That is why the anonymized sample data was requested. Without it there is not a good way to what feature/field/etc. that Protect is using.

cmsimike:
@AngellusMortis I sent you an email around Sep 3. Felt that was the best given the information has fields like `authUserId` and `accessKey`
I can send again via email if you'd like. Looking at what I have didn't have anything else around `color` other than the `isColorNightVisionEnabled` in the `ispSettings`

cmsimike:
I need to figure out signing as well but I'll work on that for now,

AngellusMortis:
No, you can send it right here in this thread. If you sent it to my email, it was probably blocked automatically.

cmsimike:
[tests.zip](https://github.com/AngellusMortis/pyunifiprotect/files/13327116/tests.zip)

AngellusMortis:
Can you generate one (or two) more with you turning on/off the color night vision and attaching/detaching the enhancer? You can  the `-w` option to make it wait longer for the Websocket messages to give you more time to do it all.

cmsimike:
I believe this capture had me turning on/off the color night vision, but I'll try again tomorrow.

AngellusMortis:
The sample_ws_messages.json is the one that is important

cmsimike:
Understood.

cmsimike:
[sample_ws_messages.zip](https://github.com/AngellusMortis/pyunifiprotect/files/13328750/sample_ws_messages.zip)
 Just the `sample_ws_messages.json` file. I was unable to remove/attach the enhancer as the cameras are installed already and I can't access them. I did spend a majority of the 60 second wait just toggling the value in the UI.

AngellusMortis:
That confirms the ISP setting is all that changes, which is to be expected, but it does not really confirm how Protect detects that a camera _has_ color night vision. 

I did more digging looking at your cameras specifically and then cross referenced them to the server app for Protect. You should be able to add a computed property to the camera class of the following to test if has color night vision:

```
@property
def has_color_night_vision(self) -> bool:
     return self.feature_flags.hotplug and self.feature_flags.hotplug.extender and self.feature_flag.hotplug.extender.is_attached
```

Then you can add the restriction back to `set_color_night_vision` using that property and HA can key off the property to decide to add the entity.

cmsimike:
I added all that we discussed (still working out the signed commit thing) but I pulled in upstream/master into this branch to resolve a conflict. VSCode's dev containers and tests are failing now (where as they were passing before I pulled upstream/master). I don't believe the failing tests are due to anything I introduced based on the logs. 

Let me know how to proceed.

AngellusMortis:
I am working on upgrading the tooling/configs for the project. Give me a bit to finish it.

AngellusMortis:
You should be able to rebase/merge and get everything to pass now. Ruff is still disabled in CI because there are a lot of failures, but I can work on those separately and it should not block you.

cmsimike:
Tests are passing for me locally 
image

cmsimike:
Opened https://github.com/AngellusMortis/pyunifiprotect/pull/319 instead in hopes the tests are fixed.

PR #325, opened 2023-12-05
Fixes broken UUID from Protect

AngellusMortis:
Protect is emitting an invalidly formatted UUID that is causing Python's native UUID type to throw an exception. This should handle the edge case.

### DIFF
--- a/pyunifiprotect/utils.py
+++ b/pyunifiprotect/utils.py
@@ -81,6 +81,7 @@ _LOGGER = logging.getLogger(__name__)
 RELEASE_CACHE = Path(__file__).parent / "release_cache.json"
 
 _CREATE_TYPES = {IPv6Address, IPv4Address, UUID, Color, Decimal, Path, Version}
+_BAD_UUID = "00000000-0000-00 0- 000-000000000000"
 
 IP_TYPES = {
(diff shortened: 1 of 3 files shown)


pyunifiprotect/utils.py
```diff
@@ -230,6 +231,10 @@ def convert_unifi_data(value: Any, field: ModelField) -> Any:
         if type_ == datetime:
             return from_js_time(value)
         if type_ in _CREATE_TYPES or (isclass(type_) and issubclass(type_, Enum)):
+            # handle edge case for improperly formated UUIDs
+            # 00000000-0000-00 0- 000-000000000000
+            if type_ == UUID and value == _BAD_UUID:
+                value = "0" * 32
```
bdraco, line 237: This one is a bit of the WTF on UI's part.   I guess we could value.replace(" ","") as well...

bdraco (approved):

pyunifiprotect/utils.py
AngellusMortis: The spaces are supposed to be zeroes though. There are only 30 zeroes in the broken string.

bdraco: Right, thats even worse 🤦

PR #335, opened 2023-12-30
Fixes some edge cases

AngellusMortis:
* Reduces the amount of events that are pulled by default to only 1 hour worth
* Do not try to cast a type that is already the correct type
* Do not fail generating sample data if smart track is missing
* Handle Protect bug where sometimes `start`/`end` query parameters are not applied correctly, emit warning to user that their query is also taking longer then expected
* More logging for handling events
* Fixes edge case where user's email can be `None`


bdraco:
Tested on one production system and didn't observe any issues or regressions.

Trying another one now

bdraco:
Screenshot 2023-12-30 at 2 44 23 PM
Screenshot 2023-12-30 at 2 44 19 PM


Saw some unexpected places on the profile but I don't think they are new

AngellusMortis:
That looks like it is overall better though since it should not be calling `get_events` if the WS is working.

bdraco:
I think we can make `bootstrap_models` a tuple so it doesn't get recreated every time its called

bdraco (approved):
Testing looks good on the 2nd system as well

bdraco:
Trying the tuple change now. If it works I'll send a PR in a moment

PR #336, opened 2023-12-31
Make bootstrap_models a tuple

bdraco:
tuples are stored as constants in the code object so they do not get recreated every time this function is called. It would have been faster to decorate it with `@cached_property` and make it a property except `@cache_property` has an undocumented class instance lock which would have made it not viable until they fixed it in py 3.12 (maybe not a problem in this instance, but I wanted to go for a simple change here)


AngellusMortis (approved):

bdraco:
thanks

PR #339, opened 2024-01-10
Updates for Protect 2.11.13

AngellusMortis:
Just a first pass because I have to expose all of these on objects as well. 

Updates for UniFi Protect 2.11.13:
* Adds `retention_duration`, `smart_detect_post_padding`, `smart_detect_pre_padding` recording settings for cameras
* Adds `audio_style` and `has_vertical_flip` feature flags for cameras
* Adds `audio_settings` for cameras
* Adds `detected_thumbnails` support for Events
* Adds `last_device_fw_updates_checked_at` for NVR
* Adds new event and smart detection enums (will be exposed on camera objects in follow up)

### DIFF
--- a/pyunifiprotect/data/nvr.py
+++ b/pyunifiprotect/data/nvr.py
@@ -62,6 +62,8 @@ if TYPE_CHECKING:
 _LOGGER = logging.getLogger(__name__)
 MAX_SUPPORTED_CAMERAS = 256
 MAX_EVENT_HISTORY_IN_STATE_MACHINE = MAX_SUPPORTED_CAMERAS * 2
+DELETE_KEYS_THUMB = {"color", "vehicleType"}
+DELETE_KEYS_EVENT = {"deletedAt", "category", "subCategory"}
 
 
 class NVRLocation(UserLocation):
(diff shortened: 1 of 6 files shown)


pyunifiprotect/data/nvr.py
```diff
@@ -124,6 +124,44 @@ class LicensePlateMetadata(ProtectBaseObject):
...
+    vehicle_type: Optional[EventThumbnailAttribute] = None
+
+    def unifi_dict(
+        self,
+        data: Optional[dict[str, Any]] = None,
+        exclude: Optional[set[str]] = None,
+    ) -> dict[str, Any]:
+        data = super().unifi_dict(data=data, exclude=exclude)
+
+        delete_keys = {"color", "vehicleType"}
```
bdraco, line 143: Would be nice to make this a constant so it doesn't get recreated every time.  Might matter since events are high volume

pyunifiprotect/data/nvr.py
```diff
@@ -253,15 +293,10 @@ def unifi_dict(
...
-        if "deletedAt" in data and data["deletedAt"] is None:
-            del data["deletedAt"]
-
-        # category/subCategory optionally added
-        if "category" in data and data["category"] is None:
-            del data["category"]
-
-        if "subCategory" in data and data["subCategory"] is None:
-            del data["subCategory"]
+        delete_keys = {"deletedAt", "category", "subCategory"}
```
bdraco, line 296: as above

bdraco (approved):

PR #341, opened 2024-01-10
Adds leak sensor helper

AngellusMortis:
`leakDetectedAt` is reset to `null` when it is no longer detecting a leak. That also means you cannot have a `last detected leak` value as well.

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -2478,6 +2478,10 @@ class Sensor(ProtectAdoptableDeviceModel):
     def is_humidity_sensor_enabled(self) -> bool:
         return self.mount_type != MountType.LEAK and self.humidity_settings.is_enabled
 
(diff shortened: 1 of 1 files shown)


bdraco (approved):

pyunifiprotect/data/devices.py
```diff
@@ -2478,6 +2478,10 @@ def is_temperature_sensor_enabled(self) -> bool:
     def is_humidity_sensor_enabled(self) -> bool:
         return self.mount_type != MountType.LEAK and self.humidity_settings.is_enabled
 
+    @property
+    def is_leak_sensor_enabled(self) -> bool:
+        return self.mount_type == MountType.LEAK
```
bdraco, line 2483:
```suggestion
        return self.mount_type is MountType.LEAK
```

If this is always an enum, identity checks are faster

PR #344, opened 2024-01-18
add ptz auto tracking

Julian0021:
The HTTP request for turning PTZ Auto Tracking simply modifies the smart detect settings of the cam.

For turning on:
  "body": "{\"smartDetectSettings\":{\"autoTrackingObjectTypes\":[\"person\"]}}",
  "method": "PATCH",

For turning off:
  "body": "{\"smartDetectSettings\":{\"autoTrackingObjectTypes\":[]}}",
  "method": "PATCH",

I dont know why its done like this, maybe they plan on enabling auto tracking on other objects than persons in the future. 
Ive added the simple functionallity to control auto track with the cli.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -118,6 +118,7 @@ unifi-protect cameras DEVICE_ID COMMAND
 | Command    | Description          |
 | ---------- | -------------------- |
 | `list-ids` | Requires no device ID. Prints list of "id name" for each device. |
+| `set-person-track` | Requires device ID. Sets person auto tracking on or off for PTZ cameras. |
 
 ##### Examples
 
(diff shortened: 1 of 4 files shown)


AngellusMortis:
There needs to be a check of whether or not the camera supports autotracking. Otherwise, there is no way to tell if a camera should have the toggle added or not.

Julian0021:
It would suffice to check if the camera is_ptz, since only ptz cams support it.

AngellusMortis:
That looks good, but it still needs to be renamed to person tracking instead of auto-tracking. In case they add vehicle/etc. in the future.

AngellusMortis:
You can use the scripts in `.bin` to run the linters and formatters. Or just to get the commands to need to run them to make CI happy.

AngellusMortis (changes requested):
Last thing, the new command should be added to the docs and a test added.

AngellusMortis (approved):

PR #349, opened 2024-02-02
Use URL objects for aiohttp requests

bdraco:
Passing URL objects avoids having to have aiohttp create them each request Its a tiny speed up to getting camera snapshots

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -260,7 +260,7 @@ class BaseApiClient:
         for attempt in range(2):
             try:
                 req_context = session.request(
-                    method, request_url, headers=headers, **kwargs
+                    method, request_url, headers=headers, **kwargs,
                 )
                 response = await req_context.__aenter__()
 
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -259,7 +259,9 @@ class BaseApiClient:
 
         for attempt in range(2):
             try:
-                req_context = session.request(method, request_url, headers=headers, **kwargs)
+                req_context = session.request(
+                    method, request_url, headers=headers, **kwargs
+                )
                 response = await req_context.__aenter__()
 
                 self._update_last_token_cookie(response)
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -177,6 +177,10 @@ class BaseApiClient:
         if session is not None:
             self._session = session
 
+        self._update_url()
+
+    def _update_url(self) -> None:
+        """Updates the url after changing _host or _port."""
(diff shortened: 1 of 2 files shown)


pyunifiprotect/api.py
```diff
@@ -176,11 +177,12 @@ def __init__(
         if session is not None:
             self._session = session
 
-    @property
-    def base_url(self) -> str:
```
AngellusMortis, line 180: Removing this is a breaking change.

AngellusMortis: Nevermind, I see you are setting it statically below. Ignore me.

pyunifiprotect/api.py
```diff
@@ -248,16 +250,18 @@ async def request(
         if require_auth:
             await self.ensure_authenticated()
 
-        url = urljoin(self.base_url, url)
+        request_url = self._url.joinpath(url[1:])
```
bdraco, line 253: This wants the relative path so we have to trim the leading `/`

tests/test_api.py
```diff
@@ -161,7 +161,7 @@ def test_base_url(protect_client: ProtectApiClient):
 
     protect_client._port = 443
 
-    assert protect_client.base_url == "https://127.0.0.1"
```
AngellusMortis, line 164: We should either add a handler for `_port`/`_host` so if they are updated somehow, `base_url` is updated, or we should remove them completely (and by extension probably this test).

bdraco: I don't see much of a use case for changing them later so I'll remove them and the test

bdraco: index = self.bootstrap.nvr.hosts.index(self._host)

Looks like we can't remove it

tests/test_api.py
```diff
@@ -160,6 +160,7 @@ def test_base_url(protect_client: ProtectApiClient):
     assert protect_client.ws_url == f"wss://127.0.0.1:0{arg}"
 
     protect_client._port = 443
+    protect_client._update_url()
```
bdraco, line 163: Since we are reaching inside the object to change to 443 I figure this is ok as well

tests/test_api.py
AngellusMortis: You could store the result of that call every time the bootstrap if refreshed. Convert the yarl obj back to just the str host, get the index and store it locally.

bdraco: Its only the snapshot urls that get called with any frequency so I left it as-is

AngellusMortis: thumbnails are also called quite a lot if you load the media browser

AngellusMortis (approved):

PR #363, opened 2024-03-20
Add persistent sessions

AngellusMortis:
Something I have wanted to support for a file since UniFi OS gets funky and returns 401 as a rate limit sometimes. 

* Adds `platformdirs` for getting the correct user cache / config directories
* Stores and loads `TOKEN` cookies to disk so `ProtectApiClient` can have persistent sessions. If you run `unifi-protect` 5-10 times rapidly, it will not randomly get a 401 since it does not issue a login each time. 
* Refactors how headers are handed so `cookie` / `x-csrf-token` are not randomly accidently deleted
* Fixes deprecation warning in CLI interface (replace `asyncio.get_event_loop` with `asyncio.run` for newer versions of Python)
* Set `x-csrf-token` on every request

This will require a HA side change to redirect the config directory to correct location.

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -491,7 +491,11 @@ class BaseApiClient:
         """Update the last token cookie."""
 
         csrf_token = response.headers.get("x-csrf-token")
-        if csrf_token is not None and csrf_token != self.headers.get("x-csrf-token"):
+        if (
+            csrf_token is not None
+            and self.headers
+            and csrf_token != self.headers.get("x-csrf-token")
+        ):
             self.set_header("x-csrf-token", csrf_token)
             await self._update_last_token_cookie(response)
 
@@ -528,7 +532,7 @@ class BaseApiClient:
         config["sessions"][session_hash] = {
             "metadata": dict(cookie),
             "value": cookie.value,
-            "csrf": self.headers.get("x-csrf-token"),
+            "csrf": self.headers.get("x-csrf-token") if self.headers else None,
         }
 
         async with aiofiles.open(self.config_file, "wb") as f:
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -440,6 +440,11 @@ class BaseApiClient:
 
     async def ensure_authenticated(self) -> None:
         """Ensure we are authenticated."""
+
+        # load session first before checking for auth
+        if self._session is None:
+            await self.get_session()
+
         if self.is_authenticated() is False:
             await self.authenticate()
 
@@ -471,16 +476,17 @@ class BaseApiClient:
                 "cookie": response.headers.get("set-cookie", ""),
             }
 
-            csrf_token = response.headers.get("x-csrf-token")
-            if csrf_token is not None:
-                self.headers["x-csrf-token"] = csrf_token
-
             self._is_authenticated = True
             await self._update_last_token_cookie(response)
             _LOGGER.debug("Authenticated successfully!")
 
     async def _update_last_token_cookie(self, response: aiohttp.ClientResponse) -> None:
         """Update the last token cookie."""
+
+        csrf_token = response.headers.get("x-csrf-token")
+        if csrf_token is not None:
+            self.headers["x-csrf-token"] = csrf_token
+
         if (
             token_cookie := response.cookies.get("TOKEN")
         ) and token_cookie != self._last_token_cookie:
@@ -544,9 +550,13 @@ class BaseApiClient:
         for key, value in session.get("metadata", {}).items():
             cookie["TOKEN"][key] = value
 
+        cookie_value = str(cookie["TOKEN"]).replace("Set-Cookie: ", "")
         self._last_token_cookie = cookie["TOKEN"]
         self._last_token_cookie_decode = None
         self._is_authenticated = True
+        self.headers = {
+            "cookie": cookie_value,
+        }
         return cookie
 
     def is_authenticated(self) -> bool:
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -246,7 +246,8 @@ class BaseApiClient:
             if force:
                 if self._session is not None:
                     self._session.cookie_jar.clear()
-                self.headers = None
+                self.set_header("cookie", None)
+                self.set_header("x-csrf-token", None)
 
             await self.ensure_authenticated()
             return self.headers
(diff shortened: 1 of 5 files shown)


pyunifiprotect/api.py
```diff
@@ -174,6 +190,10 @@ def __init__(
         self._verify_ssl = verify_ssl
         self._ws_timeout = ws_timeout
 
+        self.config_dir = config_dir or (Path(user_config_dir()) / "ufp")
+        self.cache_dir = cache_dir or (Path(user_cache_dir()) / "ufp_cache")
```
AngellusMortis, line 194: It does not seem like these do disk IO: https://github.com/platformdirs/platformdirs/blob/main/src/platformdirs/unix.py

pyunifiprotect/api.py
```diff
@@ -439,26 +480,88 @@ async def authenticate(self) -> None:
...
+            "value": cookie.value,
+        }
+
+        async with aiofiles.open(self.config_file, "wb") as f:
+            await f.write(orjson.dumps(config, option=orjson.OPT_INDENT_2))
+
+    async def _read_auth_config(self) -> SimpleCookie | None:
+        """Read auth cookie from config."""
+
+        if not await aos.path.isfile(self.config_file):
```
bdraco, line 534: Path.isfile does a blocking stat

AngellusMortis: That is why it is using aiofile instead of os.

bdraco: Sorry I missed it was wrapped in aiofiles

pyunifiprotect/api.py
```diff
@@ -439,26 +480,88 @@ async def authenticate(self) -> None:
...
+        if not session:
+            _LOGGER.debug("No existing session for %s", session_hash)
+            return None
+
+        cookie = SimpleCookie()
+        cookie["TOKEN"] = session.get("value")
+        for key, value in session.get("metadata", {}).items():
+            cookie["TOKEN"][key] = value
+
+        cookie_value = str(cookie["TOKEN"]).replace("Set-Cookie: ", "")
```
bdraco, line 558: This would break if `Set-Cookie: ` changes case

AngellusMortis: It is core functionality of `SimpleCookie` so it is not likely to change. I cannot use `lower` here because that would break the JWT. I could change it to a case insensitive regex match, but that is slower.

pyunifiprotect/api.py
```diff
@@ -439,26 +480,88 @@ async def authenticate(self) -> None:
...
+            await f.write(orjson.dumps(config, option=orjson.OPT_INDENT_2))
+
+    async def _read_auth_config(self) -> SimpleCookie | None:
+        """Read auth cookie from config."""
+
+        if not await aos.path.isfile(self.config_file):
+            _LOGGER.debug("no config file, not loading session")
+            return None
+
+        async with aiofiles.open(self.config_file, "rb") as f:
```
bdraco, line 538: Not that it likely matters here, but the common case here is that the file will exist so skipping the stat and trying to open the file (and handling the exception if it doesn't) will save one executor job which would reduce the latency here

pyunifiprotect/api.py
bdraco: `expanduser` calls `pwd.getpwuid` and `pwd.getpwnam` both of which do blocking I/O to read /etc/passwd, etc

pyunifiprotect/api.py
bdraco: I think its fine. Only pointing it out as due-diligence

bdraco:
Testing this one prod now

pyunifiprotect/api.py
AngellusMortis: `/etc/passwd` is likely going to be heavily cached by the OS those because it is an incredibly high access file, so I do not really expect that to be an issue. It still has to take the hit somewhere though. Either when ProtectApiClient is initialized or when it is imported, both are likely to be inside of an event loop.

pyunifiprotect/api.py
```diff
@@ -597,7 +699,6 @@ def __init__(
         ignore_stats: bool = False,
         ignore_unadopted: bool = True,
         debug: bool = False,
-        cache_dir: Optional[Path] = None,
```
bdraco, line 600: I guess we still need to accept a `cache_dir` and `config_dir` here so we can pass it to the super so HA can override the location

AngellusMortis: It was moved to the base class.

bdraco: Is the plan to no longer use this class in HA?

```
2024-03-21 08:19:31.109 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Maui UNVR for unifiprotect
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 549, in async_setup
    result = await component.async_setup_entry(hass, self)                                                                                                                                                                                                
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                
  File "/usr/src/homeassistant/homeassistant/components/unifiprotect/__init__.py", line 67, in async_setup_entry
    protect = async_create_api_client(hass, entry)                                                                                                                                                                                                        
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                        
  File "/usr/src/homeassistant/homeassistant/components/unifiprotect/utils.py", line 136, in async_create_api_client
    return ProtectApiClient(                                                                                                                                                                                                                              
           ^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                              
TypeError: ProtectApiClient.__init__() got an unexpected keyword argument 'cache_dir'  
```

AngellusMortis: Whoops. I guess there is not a kwargs catch on there.

AngellusMortis:
Added the kind of talked about suggestions and also added csrf tokens so they get stored as well. Otherwise, if you do a post right away (like via the CLI), it could cause a 403.

pyunifiprotect/api.py
AngellusMortis: Thinking more about it, I think the regex might be a bit faster since it does not need to scan the whole string (`^` modifier).

bdraco:
Pulling and re-testing now

bdraco (approved):

PR #367, opened 2024-03-22
Allow restoring use_global flag in set_privacy

adamjernst:
I toggled the privacy mode for my camera on and then off again in Home Assistant. I was mystified to find that it subsequently stopped recording when it had been recording before.

Turns out the camera previously had `use_global` set to `True`, so that it used the global settings for recording schedule. Changing its recording mode to `NEVER` disabled `use_global`, and it was not correctly restored later.

Add a parameter that will allow a PR in Home Assistant to restore the previous `use_global` flag.

### DIFF
--- a/pyunifiprotect/data/devices.py
+++ b/pyunifiprotect/data/devices.py
@@ -2344,6 +2344,7 @@ class Camera(ProtectMotionDeviceModel):
         enabled: bool,
         mic_level: Optional[int] = None,
         recording_mode: Optional[RecordingMode] = None,
+        use_global: Optional[bool] = None,
     ) -> None:
         """Adds/removes a privacy zone that blacks out the whole camera"""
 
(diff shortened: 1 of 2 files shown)


AngellusMortis:
This PR is not really adequate. The global recording settings have kind of been ignored completely so far in the library (I honestly kind of forgot about them). 

Before anything could be done in HA, Global recording settings would need to be fully implemented. That means:

* Simple true/false helper to turn on and off global recording settings
* Any similar helpers on Camera objects added to the NVR object to adjust global recording settings
* All Camera helpers that now overlap with global recording settings need to raise exceptions if you attempt to change them while global recording settings are enabled for a camera
* This method (`set_privacy`) should only accept one of the other: changing the recording settings manually + setting global recording to false or changing the global recording to true

AngellusMortis:
Started working on adding global recording settings support. 

First PR (https://github.com/AngellusMortis/pyunifiprotect/pull/379) handles the first two bullet points I mentioned above.

AngellusMortis:
1, 3, and 4 handled by https://github.com/AngellusMortis/pyunifiprotect/pull/380

AngellusMortis:
And go figure. 4.0.x for UniFi Protect removes global recording settings from the UI in favor of the recording manager. As far as I can tell, you can no longer use global recording settings in Protect 4.0.

PR #383, opened 2024-05-06
Support Unifi OS 4.x

Nixon506E:
Provide login fallback to new token name in UnifiOS 4.x

### DIFF
--- a/pyunifiprotect/api.py
+++ b/pyunifiprotect/api.py
@@ -498,7 +498,7 @@ class BaseApiClient:
             await self._update_last_token_cookie(response)
 
         if (
-            token_cookie := response.cookies.get("TOKEN")
+            token_cookie := response.cookies.get("TOKEN") or response.cookies.get("UOS_TOKEN")
         ) and token_cookie != self._last_token_cookie:
             self._last_token_cookie = token_cookie
             if self.store_sessions:


AngellusMortis (changes requested):
All places `TOKEN` is used has to be updated. There also needs to be a mechanism in place to determine if it is UniFi OS 4.x or if it is older and only try to load/save the correct token. Mostly for the saving/loading of sessions to disk. A UniFi OS 3.x session should not try to load if the console is running 4.x.

jasperslits:
Definitely needs more work than just the one-liner. I created https://github.com/jasperslits/pyunifiprotect/commit/dcc8141ac741663a838c39cddbb1c503073e0758 to support both cookienames and deals with the caching saving/loading. 

Now it shows a few deletions related to ptz  as 5.2.0 of this package hasnt been merged in Home Assistant yet. 

I can make a proper PR out of this but it's easier to do this after 5.2.0 is merged. The second step would be the fix for the websocket authentication.

AngellusMortis:
> I can make a proper PR out of this but it's easier to do this after 5.2.0 is merged. The second step would be the fix for the websocket authentication.

HA will likely never be updated with 5.2.0. I am waiting for pydantic 1.10.16 (https://github.com/pydantic/pydantic/pull/9162) or 4.0 to go to GA before making another version. Which ever happens first. So, if you want to get the auth updated, you probably want to do it sooner rather than later.

jasperslits:
I am about to create a proper PR for Unifi OS 4 support but looking for a bit of guidance on what works best: 
* I pull current main (5.2.0) from pyunifiprotect
* I base my changes on it
*.I submit a PR

Beause current HA - which I used - doesn't have the PTZ and other changes from 5.2.0 so it ends up with deletes. Happy to follow your guidance as I dont want to create unnecessary work when it comes to reviewing or merging the PR.

PR #388, opened 2024-05-11
Handle new cookie name for UniFi OS v4

jasperslits:
PR created against main branch for Unifi OS 4 compatibility where a change in authentication Cookie name from TOKEN to UOS_TOKEN was done. 

The PR supports both TOKEN and UOS_TOKEN cookies.


AngellusMortis:
This makes sure the cookie name is set right, but it does not make sure sessions for v3 are not loaded on v4 or vice versa. 

You will probably need to find an endpoint that allows you to figure out if it is v3 vs. v4 and load that to check _before_ setting the cookie name. `/api/system` the only unauthenticated endpoint I know of, but on v3 it does not have anything to indicate version.

```json
{
    "hardware": {
        "shortname": "UNVRPRO"
    },
    "name": "",
    "mac": "",
    "directConnectDomain": null,
    "deviceState": "setup",
    "deviceErrorCode": null,
    "uidb": {
        "guid": "",
        "id": "",
        "images": {
            "default": "7478abf0ea9850c0100338b0cd59fc80"
        }
    },
    "cloudConnected": false,
    "isOrHasGatewayInLAN": true,
    "remoteAccessEnabled": false,
    "hasInternet": true,
    "hasInternetLastChanged": 1715324581238,
    "isSingleUser": false,
    "isSsoEnabled": false
}
```

jasperslits:
So that's for a scenario where .storage/unifiprotect/unifi_protect.json is created for V3 and has a session expire date in the future, e.g. 5 days. User then upgrades from v3 to v4 and that would then use the wrong cached token?

AngellusMortis:
They do not expire in just 5 days. It is more like 30 days.

jasperslits:
So perhaps something like this would be sufficient _read_auth_config?

```Python
cookie_name = session.get("cookiename")
if cookie_name == None:
   return None
```

That would stop reading the session from the JSON and retrigger auth which in turns should write an update with cookie_name present in the JSON so we know if TOKEN vs UOS_TOKEN is to be used.

PR #397, opened 2024-05-23
Use new download mechanism to retrieve video

cyberpower678:
video/export is getting increasingly unreliable with each update of Unifi Protect.

This implements prepare/download, which is what the current UI uses, and falls back on export, to try and grab the video.

Tested for backwards compatibility, and it appears to be doing the job quite nicely.


AngellusMortis (changes requested):
api.py should be a 1:1 mapping of endpoints. So there should be two new and additional methods to prepare and download the file.

Selection and fallback logic should appear at a higher level.

cyberpower678:
> api.py should be a 1:1 mapping of endpoints. So there should be two new and additional methods to prepare and download the file.
> 
> Selection and fallback logic should appear at a higher level.

Fair enough, should I make the get_camera_video the high level method, and move prepare/download/export to new methods?

cyberpower678:
Well that was somewhat painful, but how's this?

cyberpower678:
@AngellusMortis Good morning.  Don't mean to nag, but is there anything holding up the merging of this pull request?  video/export, video/prepare, and video/download each have their own methods now.

PR #402, opened 2024-06-08
Fix cookies not getting cached properly

ep1cman:
This PR addresses two issues:

- Firstly the python standard library does not understand the `partitioned` cookie flag. This causes pyunifiprotect to not cache the session cookie, which in turn leads to errors about caused by authentication attempts exceeding rate limits. Support for this will be added in python 3.13 (https://github.com/python/cpython/issues/112713) This is fixed by modifying the standard library at run time for older versions. 
  - I could not verify this on the `3.13` beta at this time, because `PyO3`, a dependancy of pydantic fails to build on this version.
- Secondly, it saves the cookie name to the session cache because the heuristic of the version number starting with `4` proved incorrect. In my setip I have `4.0.5` on a UDM Pro, but it still uses the old `TOKEN` name.

### DIFF
--- a/pyunifiprotect/session.py
+++ b/pyunifiprotect/session.py
@@ -98,7 +98,6 @@ class SessionCache:
 
     async def read_session(
         self,
-        cookie_name: str,
         session_hash: str,
     ) -> SessionDict | None:
         """Read session cache and get specific session."""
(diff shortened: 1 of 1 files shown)


AngellusMortis (changes requested):
The cookie name cannot be stored on disk. There was a reason I removed it. If there is a significant enough change to the session that a cookie name is changed, the user needs to be forced to log in again instead of loading the existing session. The existing session may case unexpected issues.

If there is another breaking change within the same major version of UniFi OS, then `get_version_prefix` needs to be updated to handle it. I was using `/api/system` because it was an easy way to get to determine if it was newer then v4.x. But a new URL may need to be found. It does not even need to be a version number, just a prefix so you can determine which cookie name to use. 

However, with all of that since, if 4.0.3 is now the _only_ version that uses a cookie name of `UOS_TOKEN`, the best solution might just be to not support 4.0.3 since it was always only an EA version and EA versions are not really supported. This is one of the big reasons why. Ubiquiti does a shit job verifying their code still works the way it did in the previous version, and often these types of changes never even make it into stable versions. I suspect they have little to no automated testing because they have let some really big and stupid changes through (like generating a UUID with spaces instead of zeroes).

In that case, just change everything to always using `TOKEN` instead. It will make everything much simpler.

ep1cman:
Hmm thats a good point. I wouldn't trust them that this isn't a regression on their part once again though. I shall modify the check to be more specific to 4.0.3 and then should things ever change again the infrastructure is already in place.

Would you like me to split this into two PRs, one addressing the cookie name and another addressing the issue with the partitioned flag?

AngellusMortis:
The partitioned flag is a small enough change it does not really matter. And until UniFi OS is released to stable, we have no idea what Ubiquiti is going to do. I would probably still recommend just going back to always using `TOKEN`. If they change it back again, then we can worry about finding a better way to detect the newer versions at that point. Otherwise, you could spend a ton of time implementing something for no reason.

There is a reason I made it so annoying to use EA versions in the HA integration. People who want to use EA need to expect a lot of breakage since we have no control or idea of what Ubiquiti is going to do.

